PR #31, opened 2020-07-16
Improve kernel-thrown exceptions

jpsamaroo:
Removed support for LLVM < 7
Added memcpy! and memset! intrinsics
Added more efficient memcpy! implementation for output
Added global output context support
Added exception flag and exception list
Added malloc and free hostcalls
Added execution control intrinsics
Expanded ROCModule to contain exception ring buffer
Added HSAStatusSignal for exception handling
Worked around broken exception intrinsic emission

Todo:
- [x] Fix memory access fault bug (no idea how, but it's working!)
- [x] Enable and test atomic_cmpxchg implementation
- [ ] ~Test locally with highly-concurrent kernel launches~ Don't know how!
- [x] Documentation

### DIFF
--- a/src/device/runtime.jl
+++ b/src/device/runtime.jl
@@ -15,48 +15,60 @@ function load_runtime(dev_isa::String)
     GPUCompiler.load_runtime(job)
 end
 
-#@inline exception_flag() = ccall("extern julia_exception_flag", llvmcall, Ptr{Cvoid}, ())
-
 function signal_exception()
-#=
-    ptr = exception_flag()
-    if ptr !== C_NULL
-        unsafe_store!(convert(Ptr{Int}, ptr), 1)
-        threadfence_system()
-    else
-        @rocprintf("""
-            WARNING: could not signal exception status to the host, execution will continue.
-                     Please file a bug.
-            """)
-    end
-=#
+    flag_ptr = get_global_pointer(Val(:__global_exception_flag), Int64)
+    unsafe_store!(flag_ptr, 1)
+    # TODO: Can we sync here?
+    endpgm() # stop this wavefront
     return
 end
 
 function report_exception(ex)
-#=
-    @rocprintf("""
-        ERROR: a %s was thrown during kernel execution.
-               Run Julia on debug level 2 for device stack traces.
-        """, ex)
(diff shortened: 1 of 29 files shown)


jpsamaroo:
So, funny enough, I don't actually know how to get multiple wavefronts to launch in parallel for a single kernel. If anyone can figure that out and test that this still works (or not) with a ton of wavefronts throwing exceptions, I would greatly appreciate it :smile:

jpsamaroo:
@wsphillips @NorthSailor feel free to give this a spin and a review! I'll merge this once documentation is done, unless there are any changes/objections, since this work is key to completing #32

src/device/runtime.jl
```diff
@@ -15,48 +15,59 @@ function load_runtime(dev_isa::String)
...
-    @rocprintf("""
-        ERROR: a %s was thrown during kernel execution.
-               Run Julia on debug level 2 for device stack traces.
-        """, ex)
-=#
+    # Add kernel ID and exception string to exception ring buffer
+    ring_ptr = get_global_pointer(Val(:__global_exception_ring), DevicePtr{ExceptionEntry,AS.Global})
+    ring_ptr = unsafe_load(ring_ptr)
+    our_signal = _completion_signal()
+    while (ring_value = unsafe_load(ring_ptr)).kern != 1
```
vchuravy, line 30:
```suggestion
    prev = UInt64(1)
    while  prev != UInt64(0)
        prev = atomic_cas!(convert(DevicePtr{UInt64,AS.Global}, ring_ptr), UInt64(0), our_signal)
        if prev != UInt64(0)
            ring_ptr += sizeof(ExceptionEntry)
        end
        # ring_ptr bounds check
        if ....
             break
       end
    end
    if prev == UInt64(0)
           # found an empty slot
            ex_len = string_length(ex)
            copy_buf = malloc(ex_len)
            memcpy!(copy_buf, DevicePtr{UInt8,AS.Global}(ex), ex_len)
            unsafe_store!(convert(DevicePtr{UInt64}, ring_ptr)+sizeof(UInt64), UInt64(copy_buf))
    else
      # run out of space
    end
```

PR #32, opened 2020-07-16
Reimplement ROCArray using GPUArrays

jpsamaroo:


jpsamaroo:
In the interest of having a better ROCArray ASAP, I'm going to merge this once the tests are running start to finish, even if they're failing. We'll manually skip any testsets that cause a segfault, LLVM abort, or anything else that causes the test process to hang or die abnormally.

The main feature we need to pass a bunch more of these tests is a `GPUArrays.mapreducedim` implementation. Most tests failing with "Not implemented" should be fixed by having this method defined for the ROCArray. Additionally, we'll need to wrap a bunch more of the ROCm external libraries, and do so thoroughly.

jpsamaroo:
Also @antholzer if you have any ideas about the failing batched rocFFT tests, please let me know. I'm going to assume for now that the failures are just due to problems with rocFFT itself.

antholzer:
Sorry for the late reply @jpsamaroo 

I am not sure, but this rocfft [PR#305](https://github.com/ROCmSoftwarePlatform/rocFFT/pull/305) (point 1.) might provide a fix.
The following cases are failing:

1. 2D in 3D with `region=(2,3)`
2.  2D in 4D with `region=(3,4)`
3.  **real only** 1D in 2D with `region=2`
4.  **real only** 2D in 3D with `region=(1,3)` but only the inverse
5. **real only** 2D in 4F with `region=(1,4)` but only the inverse

Of these 4. and 5. cause an access fault error ([output](https://gist.github.com/antholzer/89b772977152ed4aeb8320a6b1588560)). Sadly I do not know why.

I think, that the other cases work. Interestingly batched transformation where the single batches are ordered in memory (i.e. like it is the case for batches in machine learning) work.

PR #35, opened 2020-08-04
fix for finding HIP binaries

wsphillips:
These changes should allow the HIP library to load regardless of whether it uses the current or older naming convention.
The newer location of ROCm-provided `lld.ld` is also included in the search paths.


jpsamaroo:
Thanks!

PR #39, opened 2020-08-19
Test on Julia 1.5

jpsamaroo:


jpsamaroo:
@NorthSailor any ideas on the 1.5 failure?

jpsamaroo:
Looks like it went away with a rebase. In case we encounter it again, I was able to reproduce with `Mem.pointerinfo(Mem.alloc(agent, N)).agentOwner.handle`, which when executed like that, gives an agent handle that doesn't exist in `get_agents()`. It seems we might be racing with ROCR or something in the kernel's memory system, since reordering the operations in that expression can cause the problem to go away.

NorthSailor:
> Looks like it went away with a rebase. In case we encounter it again, I was able to reproduce with `Mem.pointerinfo(Mem.alloc(agent, N)).agentOwner.handle`, which when executed like that, gives an agent handle that doesn't exist in `get_agents()`. It seems we might be racing with ROCR or something in the kernel's memory system, since reordering the operations in that expression can cause the problem to go away.

Oh, that's really odd, especially given that `get_agents()` is called by `runtests.jl` without any arguments, which should populate the dictionary with all of the available agents. Great job on tracking that down!

PR #50, opened 2020-09-17
Switch hostcall to use constant sentinels

### LINKED ISSUE #4: Use atomics/locking in hostcall
Wavefronts could race on making the hostcall or filling the buffer

jpsamaroo:
Will also switch to using an atomic CAS to transition between sentinel states, which should hopefully allow us to close #4


jpsamaroo:
This is *finally* working! Once I've figured out what the semantics should be, I'll add some tests and merge this.

PR #72, opened 2020-11-09
rocRAND

### LINKED ISSUE #14: Implement RNGs
https://github.com/ROCmSoftwarePlatform/rocRAND

antholzer:
I added support for rocRAND.
I also added support for reshape and view (ported from [CUDA.jl/src/array.jl](https://github.com/JuliaGPU/CUDA.jl/blob/master/src/array.jl) in the second commit. Let me know if it correct the way I did it (not completely sure).

I still need to get seeding to work and the GPUArrays fallbacks fail too.

Fixes #14


src/rand/rocRAND.jl
```diff
@@ -0,0 +1,37 @@
...
+const _generator = Ref{Union{Nothing,RNG}}(nothing)
+
+function default_rng()
+    if _generator[] == nothing
+        _generator[] = RNG()
+    end
+    return _generator[]::RNG
+end
+
+function GPUArrays.default_rng()
```
jpsamaroo, line 30: Needs to dispatch on array type: https://github.com/JuliaGPU/CUDA.jl/blob/4281b1b2eceba434217189ca5e4361812f3afd0f/lib/curand/CURAND.jl#L39

src/random.jl
```diff
@@ -0,0 +1,77 @@
+# random functions that dispatch either to rocRAND or GPUArrays' generic RNG
```
jpsamaroo, line 1: Generally my preference is that everything directly under `src/` can always be loaded regardless of binary dependencies, while things under `src/blas/`, `src/rand/`, etc. will be conditionally loaded when the library is available. Could we do that with this file?

test/rocarray/random.jl
```diff
@@ -0,0 +1,66 @@
+# NOTE: tests should cover both pow2 and non-pow2 dims
+
+# in-place
+for (f,T) in ((rand!,Float32),
+              (rand!,Cuint),
+              (randn!,Float32),
+              (rand_logn!,Float32),
+              (rand_poisson!,Cuint)),
+    d in (2, (2,2), (2,2,2), 3, (3,3), (3,3,3))
+    A = ROCArray{T}(undef, d)
```
jpsamaroo, line 10: Maybe we can initialize `A` with zeros, and then test that the array is non-zero after `f(A)`?

Project.toml
```diff
@@ -33,7 +34,6 @@ FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
-Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
```
jpsamaroo, line 36: `Random` should be removed from test deps below.

src/AMDGPU.jl
```diff
@@ -123,7 +123,10 @@ function __init__()
                 librocfft !== nothing      && include(joinpath(@__DIR__, "fft", "rocFFT.jl"))
                 #librocsparse !== nothing  && include("sparse/rocSPARSE.jl")
                 #librocalution !== nothing && include("solver/rocALUTION.jl")
-                #librocrand !== nothing    && include("rand/rocRAND.jl")
+                if librocrand !== nothing
```
jpsamaroo, line 126: Nitpick: If these could be merged into a single `include`, that would be a bit more consistent.

src/array.jl
```diff
@@ -205,6 +205,82 @@ function Base.copy(X::ROCArray{T}) where T
...
+    Base.unsafe_view(Base._maybe_reshape_parent(A, Base.index_ndims(I...)), I...)
+end
+
+#FIXME: add pointer conversions
+
+## reshape
+
+# optimize reshape to return a ROCArray
+
+function Base.reshape(a::ROCArray{T,M}, dims::NTuple{N,Int}) where {T,N,M}
```
jpsamaroo, line 262: Nitpick: I prefer if we stick to 4-space indentation, even if we copy code from elsewhere.

jpsamaroo (commented):
Looking great so far! Just some initial comments from a quick glance.

src/rand/rocRAND.jl
```diff
@@ -18,20 +19,26 @@ end
...
 end
 
-function GPUArrays.default_rng()
-    if _generator[] == nothing
-        _generator[] = RNG()
+function GPUArrays.default_rng(::Type{<:ROCArray})
+    if GPUARRAY_RNG[] == nothing
+        agent = AMDGPU.default_device().device.agent
+        p = Ref{UInt32}()
+        AMDGPU.getinfo(agent, HSA.AGENT_INFO_WORKGROUP_MAX_SIZE, p) |> AMDGPU.check
```
jpsamaroo, line 36: Needs an `GC.@preserve p begin AMDGPU.getinfo(...) end`

src/AMDGPU.jl
jpsamaroo: I was thinking more that the file can be included in `src/rand/rocRAND.jl` since it depends on rocRAND. Not a big deal though, I might just change it later :slightly_smiling_face:

src/rand/interface.jl
```diff
@@ -0,0 +1,77 @@
...
+rand_logn(T::rocRAND.LognormalType, dim1::Integer, dims::Integer...; kwargs...) =
+    rocRAND.rand_logn(rocrand_rng(), T, Dims((dim1, dims...)); kwargs...)
+rand_poisson(T::rocRAND.PoissonType, dim1::Integer, dims::Integer...; kwargs...) =
+    rocRAND.rand_poisson(rocrand_rng(), T, Dims((dim1, dims...)); kwargs...)
+
+# GPUArrays in-place
+Random.rand!(A::ROCArray) = Random.rand!(gpuarrays_rng(), A)
+Random.randn!(A::ROCArray; kwargs...) =
+    error("AMDGPU.jl does not support generating normally-distributed random numbers of type $(eltype(A))")
+# FIXME: GPUArrays.jl has a randn! nowadays, but it doesn't work with e.g. Cuint
```
jpsamaroo, line 47: Is this a known issue for GPUArrays? If not, could you file one on GPUArrays?

jpsamaroo:
Is this behavior expected?
```
julia> using AMDGPU, Random

julia> A = ROCArray{Int32,1}(undef, 8)
8-element ROCArray{Int32,1}:
 -1094795586
 -1094795586
 -1094795586
 -1094795586
 -1094795586
 -1094795586
 -1094795586
 -1094795586

julia> AMDGPU.rand!(A);

julia> A
8-element ROCArray{Int32,1}:
 -238733085
 -238733085
 -238733085
 -238733085
 -238733085
 -238733085
 -238733085
 -238733085
```

src/AMDGPU.jl
antholzer: Ah, ok. But then the function defined there (e.g. `seed!`) would not be available as `AMDGPU.seed!`, or could this still be done?

jpsamaroo: If that's the case, then you could split out the code that doesn't use rocRAND directly to `src/random.jl`.

antholzer:
This behavior is not expected. I forgot to properly seed the RNGs. Should now be fixed.

The only other thing that is still missing is Int64 support.

jpsamaroo:
Yeah I see that `Int64` causes a memory fault... Do you know if this is a rocRAND issue? I didn't see anything related on their issue tracker.

antholzer:
This is an issue on our side, since random `Integer` are implemented in GPUArrays in [host/random.jl](https://github.com/JuliaGPU/GPUArrays.jl/blob/master/src/host/random.jl). The problem is probably in line 44, since `Int32` is working.

jpsamaroo:
Ok, well if you get the chance to investigate a fix then feel free, but I'm willing to merge this without the fix (especially since I still need to get us working on GPUArrays master).

src/AMDGPU.jl
antholzer: Ok, I will think about a good way to do this.
However, we then also need to tests which functions work with the GPUArrays random number generator.

How should we handle e.g. `AMDGPU.seed!`, since it should seed both rngs, if rocrand is available, and only the GPUArrays one otherwise?
Should I just define `rocrand_rng` in interface.jl to be the same as `gpuarrays_rng` if rocrand is not available?

jpsamaroo: > Should I just define `rocrand_rng` in interface.jl to be the same as `gpuarrays_rng` if rocrand is not available?

It depends on the difference in supported functionality between rocrand and gpuarrays RNG; if GPUArrays supports about the same amount of calls as rocRAND, then yes, just default to GPUArrays.

antholzer:
I now split the GPUArrays and rocrand interface such that one can use the GPUArrays functions if rocrand is not installed.
Not sure, if I did it correctly, but it seems to work.

It defaults to rocrand implementations when possible.
I did not add tests, since the one in GPUArrays should cover it (once they work).

jpsamaroo (approved):

jpsamaroo:
Tests pass and I'm happy with how this looks. I just want to test this without rocRAND available to make sure things still work.

jpsamaroo:
Thanks so much!

PR #75, opened 2020-12-11
Update to Julia 1.6

jpsamaroo:
Notably removes `DevicePtr` in favor of `Core.LLVMPtr`, which makes `@generated` functions much easier to write and probably allows the backend to optimize better.


jpsamaroo:
Currently we're hitting assertions in julia-debug due to https://github.com/JuliaLang/julia/issues/38864, and CI has been crashing for reasons beyond my understanding. I've had 99.9% of tests passing locally on a julia-debug+LLVM asserts build, though, so I'm going to merge this anyway once CI finishes.

PR #77, opened 2021-01-11
rocconvert captured values in closures

jpsamaroo:
From https://github.com/JuliaGPU/CUDA.jl/pull/625


jpsamaroo:
Implemented in #240

PR #104, opened 2021-02-19
Add mapreducedim!

jpsamaroo:
The long-awaited `GPUArrays.mapreducedim!` arrives! This implementation was copied and modified from oneAPI. We should be able to start enabling more tests, now that we have this.


vchuravy:
Fantastic!

PR #131, opened 2021-04-08
Fix copyto! on view; include offset

### LINKED ISSUE #130: "Spills" from adjacent views of ROCVector
With AMDGPU 0.2.5, I get

```
julia> function f(xs)
           fill!(xs, 0)
           a = @view xs[1:end÷2]
           b = @view xs[end÷2+1:end]
           fill!(a, 1)
           a[1], a[end], b[1], b[end], b[[1, end]]
       end
f (generic function with 1 method)

julia> f(ROCVector{Int}(undef, 2^11))
(1, 1, 1, 1, [0, 0])

julia> f(Vector{Int}(undef, 2^11))
(1, 1, 0, 0, [0, 0])
```

but I expect `b[1] == b[end] == 0`. It's also strange that `b[[1, end]]` is different from `[b[1], b[end]]`.

tkf:
fix #130

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -172,7 +172,7 @@ function Base.copyto!(dest::Array{T}, d_offset::Integer,
     @boundscheck checkbounds(source, s_offset+amount-1)
     wait!(source)
     Mem.download!(pointer(dest, d_offset),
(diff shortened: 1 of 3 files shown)


jpsamaroo:
Thanks!

Issue #139, opened 2021-06-03
Missing `active_kernels`


michel2323:
`AMDGPU.jl` misses an `active_kernels` API.

 ```julia
julia>             TestPolarFormulation.runtests(datafile, device, AT)
PolarForm API: Error During Test at /ccs/proj/csc359/michel/git/ExaPF.jl/test/Polar/TestPolarForm.jl:31
  Got exception outside of a @test
  UndefVarError: active_kernels not defined
  Stacktrace:
    [1] Event(#unused#::ROCDevice)
      @ ROCKernels /ccs/proj/csc359/michel/julia_depot_spock/packages/ROCKernels/s7LRW/src/ROCKernels.jl:82
    [2] transfer!(polar::PolarForm{Float64, ROCVector{Int64}, ROCVector{Float64}, ROCMatrix{Float64}}, buffer::ExaPF.PolarNetworkState{ROCVector{Int64}, ROCVector{Float64}}, u::ROCVector{Float64})
      @ ExaPF /ccs/proj/csc359/michel/git/ExaPF.jl/src/Polar/kernels.jl:277
```
```julia
(ExaPF) pkg> status
     Project ExaPF v0.5.0
      Status `/autofs/nccs-svm1_proj/csc359/michel/git/ExaPF.jl/Project.toml`
  [21141c5a] AMDGPU v0.2.7 `/ccs/proj/csc359/michel/julia_depot_spock/dev/AMDGPU`
  [052768ef] CUDA v3.2.1
  [6a86dc24] FiniteDiff v2.8.0
  [f6369f11] ForwardDiff v0.10.18
  [63c18a36] KernelAbstractions v0.6.0
  [ba0b0d4f] Krylov v0.6.1
  [093fc24a] LightGraphs v1.3.5
  [b8f27783] MathOptInterface v0.9.22
  [2679e427] Metis v1.0.0
  [7eb9e9f0] ROCKernels v0.1.0
  [47a9eef4] SparseDiffTools v1.13.2
  [a759f4b9] TimerOutputs v0.5.9
  [37e2e46d] LinearAlgebra
  [de0858da] Printf
  [2f01184e] SparseArrays
```

PR #141, opened 2021-06-11
Add support for active_kernels

michel2323:
Related to https://github.com/JuliaGPU/AMDGPU.jl/issues/139

### DIFF
--- a/src/AMDGPU.jl
+++ b/src/AMDGPU.jl
@@ -63,13 +63,7 @@ include(joinpath("device", "runtime.jl"))
 include(joinpath("device", "llvm.jl"))
 include(joinpath("device", "globals.jl"))
 
-const _active_kernels = IdDict{HSAQueue,Vector{AMDGPU.HSAStatusSignal}}()
-function active_kernels()
-    lock(RT_LOCK) do
-        copy(_active_kernels)
-    end
-end
-
+include("query.jl")
 include("compiler.jl")
 include("execution_utils.jl")
 include("execution.jl")
--- a/src/AMDGPU.jl
+++ b/src/AMDGPU.jl
@@ -31,6 +31,8 @@ import .HSA: Agent, Queue, Executable, Status, Signal
 
 struct Adaptor end
 
+const RT_LOCK = Threads.SpinLock()
+
 include("extras.jl")
 include("error.jl")
 include("agent.jl")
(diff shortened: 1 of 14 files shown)


src/AMDGPU.jl
```diff
@@ -153,6 +154,7 @@ function __init__()
                 Please run Pkg.build("AMDGPU") and reload AMDGPU.
                 """
             end
+            active_kernels[get_default_queue()] = Vector{AMDGPU.RuntimeEvent{AMDGPU.HSAStatusSignal}}()
```
jpsamaroo, line 157: This should move to the `HSAQueue` constructor, which will be invoked by `get_default_queue()` on its first call for a given agent.

src/execution.jl
```diff
@@ -193,6 +193,7 @@ macro roc(ex...)
                     foreach($wait!, ($(var_exprs...),))
                     if $launch
                         local $signal = $create_event($kernel.mod.exe)
+                        push!(AMDGPU.active_kernels[AMDGPU.get_default_queue()], $signal)
```
jpsamaroo, line 196: This won't work if the queue selected by the user (and passed in `call_kwargs`) isn't the default queue.

src/runtime.jl
```diff
@@ -72,6 +72,7 @@ function Base.wait(event::RuntimeEvent{HSAStatusSignal}; check_exceptions=true,
                 deleteat!(mod.metadata, idx)
             end
         end
+        deleteat!(active_kernels[get_default_queue()], findall(x->x==event, active_kernels[get_default_queue()]))
```
jpsamaroo, line 75: We should probably get the queue from the event. I'm not sure if our `HSAStatusSignal` contains the queue it was launched on, so we'll possibly have to add that as a field.

jpsamaroo (commented):
:heart:

jpsamaroo:
I'm trying to push changes to your branch, but I'm getting a permissions error. Did you enable the option to let maintainers push to your branch?

michel2323:
For some reason I don't see the "Allow edits from maintainers" option. I just gave you admin rights.

jpsamaroo:
Thanks, pushed!

jpsamaroo:
I hope you don't mind, I'm going to use this PR to rip out some old functionality that just complicates the code base for no tangible benefit.

michel2323:
Of course not!

jpsamaroo:
I've modified the API of `HSAStatusSignal` so that each signal always generates a background waiter which waits on the real HSA signal, and notifies tasks that called `wait` on the signal once that's finished. All foreground waiting tasks will also throw a `KernelException` if the kernel threw an exception.

This background task will also remove the kernel from the set of active kernels. @michel2323 I added a runtime lock, which will need to be taken to manipulate `_active_kernels`, and exposed the function `active_kernels()`, which is now the user-facing API.

jpsamaroo:
Ok, just need to pass the new test, and add docs. @michel2323 the API has been changed to accept a queue argument, and will default to the default queue.

jpsamaroo:
Thanks a ton @michel2323 !

PR #159, opened 2021-10-06
Add `has_rocm_gpu` util function

### LINKED ISSUE #158: `has_rocm_gpu`: a utility function that returns `true` if a GPU is available?
The `CUDA.has_cuda_gpu()` function is pretty useful so I was wondering if there was an AMDGPU.jl equivalent. 

Skimming the codebase I couldn't find one so I came up with this which is probably not great but thought I'd open an issue about the util function anyways.

```julia
function has_rocm_gpu()
	for agent in AMDGPU.get_agents()
		if agent.type == :gpu
			return true
		end
	end
	return false
end
```

ali-ramadhan:
Resolves #158 

Don't think it currently works on machines without ROCm since `AMDGPU.get_agents()` errors:

```
julia> AMDGPU.get_agents()
ERROR: HSA error (code #4107, HSA_STATUS_ERROR_NOT_INITIALIZED: An API other than hsa_init has been invoked while the reference count of the HSA runtime is zero.)
Stacktrace:
 [1] check
   @ ~/tmp/depot/packages/AMDGPU/Z873w/src/error.jl:34 [inlined]
 [2] |>
   @ ./operators.jl:858 [inlined]
 [3] get_agents()
   @ AMDGPU ~/tmp/depot/packages/AMDGPU/Z873w/src/agent.jl:100
 [4] top-level scope
   @ REPL[5]:1
```

but if there's a way around this then `AMDGPU.has_rocm_gpu()` should work on machines without any GPUs.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -29,5 +29,9 @@ function versioninfo(io::IO=stdout)
 end
 
 function has_rocm_gpu()
-    return length(AMDGPU.get_agents(:gpu)) > 0
+    if !AMDGPU.hsa_configured 
+        return false
+    else
+        return length(AMDGPU.get_agents(:gpu)) > 0
+    end
 end


jpsamaroo:
You can just check the `hsa_configured` variable, and if it's `false`, just return immediately before querying the agents.

PR #174, opened 2021-10-31
Add errorhandler for async queue errors

vchuravy:


jpsamaroo:
Segfaults are likely the same as reported in https://github.com/JuliaGPU/AMDGPU.jl/pull/163#issuecomment-953195867, and I suspect are due to a race in hostcall. I suspect #50 would fix them.

jpsamaroo:
We still need to ensure that we remove the dead queue from `DEFAULT_QUEUES` if it's there, and from `_active_kernels`. Also, we need to kill any running hostcalls on that queue, since otherwise they'll just hang.

src/queue.jl
```diff
@@ -6,31 +6,57 @@ Each queue is uniquely associated with an agent.
...
+    # TODO: From which thread are we calling this?
+    if status !== HSA.STATUS_SUCCESS then
+        queue = QUEUES[_queue]
+        queue.active = false
+        @assert DEFAULT_QUEUES[queue.agent] == queue # @jpsamaroo who gurantuees this 1:1 relation?
+        delete!(DEFAULT_QUEUES, queue.agent)
+        lock(RT_LOCK) do
+            delete!(_active_kernels, queue)
+        end
+        # do we need to call `HSA.queue_destroy`?
```
jpsamaroo, line 25: I'm not particularly clear on the semantics, but I do know we're not allowed to destroy the queue in the queue callback, because `queue_destroy` waits on any running callbacks to terminate. I think we should probably `@async` destroy the queue.

src/queue.jl
```diff
@@ -6,31 +6,57 @@ Each queue is uniquely associated with an agent.
...
-    queue::Ref{Ptr{HSA.Queue}}
+    queue::Ptr{HSA.Queue}
     active::Bool
 end
-
+const QUEUES = WeakKeyDict{Ptr{HSA.Queue},HSAQueue}()
 const DEFAULT_QUEUES = IdDict{HSAAgent,HSAQueue}()
 
+function queue_error_handler(status::HSA.Status, _queue::Ptr{HSA.Queue}, data::Ptr{Cvoid})::Nothing
+    # TODO: From which thread are we calling this?
```
jpsamaroo, line 16: The HSA RT docs say, "it will be invoked in an HSA runtime thread", so we should assume that we need to be careful about accessing shared globals.

src/queue.jl
```diff
@@ -6,31 +6,57 @@ Each queue is uniquely associated with an agent.
...
+    if status !== HSA.STATUS_SUCCESS then
+        queue = QUEUES[_queue]
+        queue.active = false
+        @assert DEFAULT_QUEUES[queue.agent] == queue # @jpsamaroo who gurantuees this 1:1 relation?
```
jpsamaroo, line 20: This is wrong, the default queue isn't necessarily the one calling this handler. This is managed by the `HSAAgent` ctor, IIRC, as well as the launch code, which will create a new default queue if one doesn't exist.

src/queue.jl
```diff
@@ -6,31 +6,57 @@ Each queue is uniquely associated with an agent.
...
     finalizer(queue) do queue
+        if haskey(QUEUES, queue.queue) then
+            delete!(QUEUES, queue.queue)
+        end
         if queue.active
-            HSA.queue_destroy(queue.queue[]) |> check
+            HSA.queue_destroy(queue.queue) |> check
         end
         lock(RT_LOCK) do
+            # XXX: What if someone already did this?
```
jpsamaroo, line 59: We should probably ensure that if `queue.active == false`, it's already been deleted.

src/queue.jl
```diff
@@ -58,7 +84,8 @@ end
 
 "Kills all kernels executing on the given queue, and destroys the queue."
 function kill_queue!(queue::HSAQueue)
+    # TODO (jpsamaroo): Do we need to check for queue being in any of these?
     queue.active = false
-    delete!(DEFAULT_QUEUES, queue.agent)
-    HSA.queue_destroy(queue.queue[]) |> check
+    delete!(DEFAULT_QUEUES, queue.agent)
```
jpsamaroo, line 89: Whitespace

jpsamaroo: And use `QUEUES`, not `DEFAULT_QUEUES`. I'm not sure why I did this...

src/queue.jl
```diff
@@ -58,7 +84,8 @@ end
 
 "Kills all kernels executing on the given queue, and destroys the queue."
 function kill_queue!(queue::HSAQueue)
+    # TODO (jpsamaroo): Do we need to check for queue being in any of these?
```
jpsamaroo, line 87: We should return if `queue.active == false`

src/queue.jl
```diff
@@ -6,31 +6,57 @@ Each queue is uniquely associated with an agent.
...
+    queue = HSAQueue(agent, r_queue[], true)
+    @assert !haskey(QUEUES, queue.queue)
+    QUEUES[queue.queue] = queue
 
     lock(RT_LOCK) do
         _active_kernels[queue] = Vector{AMDGPU.RuntimeEvent{AMDGPU.HSAStatusSignal}}()
     end
 
     hsaref!()
     finalizer(queue) do queue
```
jpsamaroo, line 51: Maybe we can consider moving this logic to `kill_queue!`?

src/queue.jl
```diff
@@ -6,31 +6,57 @@ Each queue is uniquely associated with an agent.
 """
 mutable struct HSAQueue
     agent::HSAAgent
-    queue::Ref{Ptr{HSA.Queue}}
+    queue::Ptr{HSA.Queue}
     active::Bool
 end
-
+const QUEUES = WeakKeyDict{Ptr{HSA.Queue},HSAQueue}()
```
jpsamaroo, line 12: To be replaced with something different, per Slack discussion.

jpsamaroo (changes requested):

jpsamaroo:
We'll probably need to generate an async error manually to test this code...

src/queue.jl
```diff
@@ -1,38 +1,75 @@
...
+        return (; old=y, success=ok)
+    end
 end
-
+const QUEUES = Dict{Ptr{HSA.Queue},WeakRef}()
 const DEFAULT_QUEUES = IdDict{HSAAgent,HSAQueue}()
 
+function queue_error_handler(status::HSA.Status, _queue::Ptr{HSA.Queue}, data::Ptr{Cvoid})::Nothing
+    # FIXME: Not libuv-safe
+    if status !== HSA.STATUS_SUCCESS then
```
jpsamaroo, line 37:
```suggestion
    if status !== HSA.STATUS_SUCCESS
```

Lol

jpsamaroo:
@JonChesterfield pointed out that `AMDGPU.trap()` in a launched kernel will trigger an async queue error, which I'm now using to trigger the callback. Unfortunately, the current callback tries to access PTLS and the GC, which won't work while on HSA's thread. I'm going to try just setting a global per-queue signal flag and adding a background listener task for each queue that will kill the queue for us (AKA "a hack").

jpsamaroo:
The latest push now properly and safely detects async queue errors, kills the queue from a Julia thread, and throws `HSAError`s on all signals waiting on a kernel that was running on the killed queue. Need some tests, and then this is good to go!

src/queue.jl
```diff
@@ -1,38 +1,119 @@
...
-mutable struct HSAQueue
-    agent::HSAAgent
-    queue::Ref{Ptr{HSA.Queue}}
-    active::Bool
+@static if VERSION >= v"1.7-"
+    """
+    An HSA queue.
+    Each queue is uniquely associated with an agent.
+    """
+    @eval mutable struct HSAQueue
```
vchuravy, line 8: Why do we need the eval here, but not below? To avoid the expansion of `@atomic`?

src/queue.jl
```diff
@@ -1,38 +1,119 @@
...
+const QUEUE_STATUS = Dict{Ptr{HSA.Queue},HSA.Status}()
+const QUEUE_STATUS_COND = Dict{Ptr{HSA.Queue},Base.AsyncCondition}()
 const DEFAULT_QUEUES = IdDict{HSAAgent,HSAQueue}()
 
+async_send(data::Ptr{Cvoid}) = ccall(:uv_async_send, Cint, (Ptr{Cvoid},), data)
+
+function queue_error_handler(status::HSA.Status, _queue::Ptr{HSA.Queue}, data::Ptr{Cvoid})::Nothing
+    if status != HSA.STATUS_SUCCESS
+        l = RT_LOCK.cond_wait.lock
+        while true
```
vchuravy, line 42: Instead of having a lock, how about each queue has a field for the associated condition? Or we pass the handle in as the data?

src/queue.jl
jpsamaroo: Yeah, vestige of pre-`@static`. Will remove.

src/queue.jl
jpsamaroo: That sounds excellent to me! Will do that with my next push.

jpsamaroo: I also need to set the error status code, so I'll just pass in the pointer to the `HSAQueue` object itself.

src/queue.jl
```diff
@@ -1,64 +1,156 @@
...
 end
-
+const QUEUES = Dict{Ptr{HSA.Queue},WeakRef}()
 const DEFAULT_QUEUES = IdDict{HSAAgent,HSAQueue}()
 
+async_send(data::Ptr{Cvoid}) = ccall(:uv_async_send, Cint, (Ptr{Cvoid},), data)
+
+function queue_error_handler(status::HSA.Status, _queue::Ptr{HSA.Queue}, queue_obj_ptr::Ptr{Cvoid})::Nothing
+    if status != HSA.STATUS_SUCCESS
+        queue = unsafe_pointer_to_objref(queue_obj_ptr)::HSAQueue
```
vchuravy, line 43: I think there is the potential risk here that the `queue` got GC'd.

jpsamaroo: If the queue gets GC'd, the finalizer will first destroy the queue. This should prevent async callbacks from firing after the finalizer completes, since destroying the queue waits for any pending callbacks to complete.

PR #176, opened 2021-11-02
rocprintf: Avoid reading bad data

jpsamaroo:


src/device/gcn/output.jl
```diff
@@ -156,6 +156,26 @@ function wave_serialized(func::Function)
     return result
 end
 
+const SC_PAGESIZE = 30
+const MS_ASYNC = 1
+@static if Sys.islinux()
+const PAGESIZE = ccall(:sysconf, Clong, (Cint,), SC_PAGESIZE)
+end
```
vchuravy, line 163:
```suggestion
if Sys.islinux()
    const PAGESIZE = ccall(:sysconf, Clong, (Cint,), SC_PAGESIZE)
end

```

src/device/gcn/output.jl
```diff
@@ -156,6 +156,26 @@ function wave_serialized(func::Function)
...
+function semi_safe_load(ptr::Ptr{T}) where T
+    num_pages = ceil(Int, sizeof(T) / PAGESIZE)
```
vchuravy, line 165: What if the object spans two pages, but it itself is `< page`?

src/device/gcn/output.jl
```diff
@@ -174,7 +194,10 @@ function Base.unsafe_load(ptr::LLVMPtr{ROCPrintfBuffer,as} where as)
         ptr += sizeof(UInt64)
         UInt64(T_ptr) == 0 && break
         T = unsafe_pointer_to_objref(T_ptr)
-        arg = unsafe_load(convert(Ptr{T}, ptr))
+        valid, arg = semi_safe_load(convert(Ptr{T}, ptr))
+        if !valid
+            @warn "@rocprintf: Memory read failed!\nFuture read failures will be ignored" maxlog=1
+        end
         push!(args, arg)
```
vchuravy, line 201:
```suggestion
        arg = arg::T
        push!(args, arg)
```

jpsamaroo: This will of course fail if we get `nothing`. I'm thinking we could return a zero-initialized object of the appropriate type (which must necessarily be bitstype), and indicate in the warning that printed data may be incorrect.

src/device/gcn/output.jl
jpsamaroo: Latest commit should handle this

vchuravy:
Does this actually fix the observed issue?

jpsamaroo:
It doesn't fix the underlying issue; this is a hotfix before I merge #50, which is also not guaranteed to fix this issue (since I'm not sure if there's another race occurring). Hence the warning.

And the observed race itself is sporadic, but I should definitely add some manual tests for this helper. Will mark as draft until then.

PR #185, opened 2021-12-16
Search for device libs relative to LD_LIBRARY_PATH

jpsamaroo:
@luraess reported that AMDGPU doesn't find bitcode files on systems using Spack-provided ROCm installs. This PR works around this (and similar setups) by searching `../amdgcn/bitcode/ocml.*` relative to all entries in `LD_LIBRARY_PATH`, which Spack and friends usually will set to point to provided libraries.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -106,6 +106,25 @@ function find_ld_lld()
     return ""
 end
 
+function find_device_libs()
+    if isdir("/opt/rocm/amdgcn/bitcode")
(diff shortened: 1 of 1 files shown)


luraess:
Thanks for looking into this issue! Current fix still errors. See [error_amdgpu.txt](https://github.com/JuliaGPU/AMDGPU.jl/files/7731025/error_amdgpu.txt) for stack trace.

PR #194, opened 2022-02-11
Function overrides lead to undefined functions in KA

michel2323:
The function overrides of `sqrt` and `abs` make only these methods available in the namespace of `AMDGPU` from `Kernelabstractions.jl`. This leads to `sqrt(::Float64)` for example missing in KA.

I assume other function overrides lead to the same behavior.


jpsamaroo:
@michel2323 since device overrides are only in #187 , I changed the base branch to `jps/julia-1.7`. Can you rebase these changes on that branch?

michel2323:
@jpsamaroo Sorry for the late reply. `compat.jl` has been removed, so the only difference is that I had to remove `abs` overrides as they would also not be available in `KA`.

jpsamaroo:
Thanks!

PR #196, opened 2022-02-16
Define a map from gamma to tgamma

vchuravy:
Like https://github.com/JuliaGPU/CUDA.jl/pull/1361

I am wondering if we could actually get rid of GCNIntrinsic, I find that style of metaprogramming
much harder to reason through.


jpsamaroo (approved):

jpsamaroo:
I personally like it, since it minimizes a lot of duplicated code. If we could get rid of the `_intr` generated function but still use some simple metaprogramming to prevent duplicating code for `(Float16, Float32, Float64)` et. al, I'd be fine with that.

jpsamaroo:
Thanks!

PR #215, opened 2022-03-22
Mem: Fix DL/UL/TX for wrapped ptr

jpsamaroo:
@luraess

Todo:
- [x] Test if download/upload is necessary for host-backed pointer (required, updates do not propagate automatically)
- [x] Fix segfaults


luraess:
Thanks! ❤️

PR #221, opened 2022-03-31
HSA: Don't always load hsa_rocr_jll

jpsamaroo:
@michel2323 this should fix the missing agents on Crusher


michel2323:
I confirm it's fixed!

PR #224, opened 2022-04-01
Restrict multi-GPU test to GPU agent only

### LINKED ISSUE #223: Restrict agent to `:gpu` in HSA agent test
In HSA agents tests, the multi-GPU list 
https://github.com/JuliaGPU/AMDGPU.jl/blob/bd8b4c400bd440c571beb33f9921be28323ce573/test/hsa/agent.jl#L11
should only get `:gpu` agents as it will otherwise error if CPU agents show up. This modification should make it:
```
if length(AMDGPU.get_agents(:gpu)) > 1 
```

luraess:
Restrict agent to `:gpu` in HSA agent test. Closes #223.

### DIFF
--- a/test/hsa/agent.jl
+++ b/test/hsa/agent.jl
@@ -8,7 +8,7 @@
         @test length(agent_name) > 0
         @test !occursin('\0', agent_name)
 
-        if length(AMDGPU.get_agents()) > 1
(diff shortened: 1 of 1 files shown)


jpsamaroo:
Thanks!

PR #232, opened 2022-05-12
Remove launch export

matinraayai:
Changed launch! export to launch to be consistent with CUDA.jl.

### DIFF
--- a/src/AMDGPU.jl
+++ b/src/AMDGPU.jl
@@ -17,7 +17,7 @@ import Core: LLVMPtr
 ### Exports ###
 
 export HSAAgent, HSAQueue, HSAExecutable, HSAKernelInstance, HSASignal
-export get_agents, profile, get_first_isa_string, launch!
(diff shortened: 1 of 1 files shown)


src/AMDGPU.jl
```diff
@@ -17,7 +17,7 @@ import Core: LLVMPtr
 ### Exports ###
 
 export HSAAgent, HSAQueue, HSAExecutable, HSAKernelInstance, HSASignal
-export get_agents, profile, get_first_isa_string, launch!
+export get_agents, profile, get_first_isa_string, launch
```
jpsamaroo, line 20:
```suggestion
export get_agents, profile, get_first_isa_string
```

Let's just get rid of this so we don't conflict at all.

matinraayai:
@jpsamaroo I committed the change. you can merge now.

jpsamaroo:
Thanks!

PR #233, opened 2022-05-22
RFC: Support Adapt.AbstractGPUDevice

oschulz:
See JuliaGPU/Adapt.jl#52

Not functional like this yet, I don't know AMDGPU well enough and don't have a device to test, currently. See JuliaGPU/CUDA.jl#1520 for the CUDA.jl counterpart.

### DIFF
--- a/src/agent.jl
+++ b/src/agent.jl
@@ -1,6 +1,6 @@
 # Utilities for working with HSA agents
 
-mutable struct HSAAgent
+mutable struct HSAAgent <: AbstractGPUDevice
     agent::HSA.Agent
 
     # Cached information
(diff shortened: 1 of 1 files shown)


jpsamaroo:
Thanks for putting this together! CI is down while I'm out of the country, so it'll have to wait another week before I can get CI back up.

oschulz:
> Thanks for putting this together! CI is down while I'm out of the country

Thanks @jpsamaroo! When you're back, I'll need some hints from you on the AMDGPU side of this (I don't really know what the equivalent of CuDevice is, and so on). :-)

src/device/array.jl
```diff
@@ -1,6 +1,23 @@
...
+
+
+# ToDo: This is not sufficient/general enough (doesn't support multiple devices)
+# What's the AMDGPU equivalent of CUDA.CuDevice?
+"""
+    struct ROCDevice <: AbstractGPUDevice
+
+A ROC device.
+"""
+struct ROCDevice <: AbstractGPUDevice end
```
jpsamaroo, line 13: Can you instead change https://github.com/JuliaGPU/AMDGPU.jl/blob/99967b75deb4d78c12236e9a77943b4e0294af35/src/agent.jl#L3 to `mutable struct HSAAgent <: AbstractGPUDevice`? We use `HSAAgent` for most operations involving devices.

oschulz: Thanks, done! How do I do

```julia
Adapt.adapt_storage(device::HSAAgent, x) = adapt_storage(ROCArray, x)
```

correctly (i.e. send the array to the right device)?

jpsamaroo:
@oschulz can you rebase this on AMDGPU master? `HSAAgent` is now named `ROCDevice`.

oschulz:
Will do, I need to rebase the other related PRs as well anyway.

oschulz:
> @oschulz bump

Sorry for the delay, it's not forgotten, but holidays got in the way. :-) Also looks like the whole thing will now include ArrayInterfaceCore as well (recent discussion on JuliaGPU/Adapt.jl#52). Gimme a bit of time to adapt to that, I'll modify and rebase the whole PR-complex in the process.

oschulz:
Closed, but not forgotten, to be reincarnated in the future (see JuliaGPU/Adapt.jl#64).

PR #246, opened 2022-07-18
Add BackToCPU struct to reduce 'view' allocations

pxl-th:
This PR adopts the same approach for view indices as in [CUDA.jl](https://github.com/JuliaGPU/CUDA.jl/blob/master/src/array.jl#L659) to reduce allocations.

Before:

```julia
julia> x = AMDGPU.rand(Float32, 1024, 1024);
julia> @time @view(x[1:1024, 1:1024]);
  0.000010 seconds (3 allocations: 16.312 KiB)
```

This PR:

```julia
julia> x = AMDGPU.rand(Float32, 1024, 1024);
julia> @time @view(x[1:1024, 1:1024]);
  0.000004 seconds (1 allocation: 64 bytes)
```

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -250,12 +250,18 @@ rocviewlength() = ()
 @inline rocviewlength(i1::AbstractUnitRange, I...) = (length(i1), rocviewlength(I...)...)
 @inline rocviewlength(i1::AbstractUnitRange, ::Base.ScalarIndex...) = (length(i1),)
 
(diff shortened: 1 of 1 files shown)


jpsamaroo (approved):
Awesome! I'm going to pull this into my local branch and validate that everything still works, in the absence of working CI, before I merge.

jpsamaroo:
Thanks!

PR #251, opened 2022-07-28
Mem: Retry failing allocations

jpsamaroo:
This automatically retries HSA allocations up to 3 times if they initially fail with an `OUT_OF_RESOURCES` code, with calls to the GC in-between to try to free up allocated objects.

@claforte @pxl-th


claforte:
Thanks a lot Julian! @pxl-th did this resolve all your issues?

pxl-th:
> Thanks a lot Julian! @pxl-th did this resolve all your issues?

Some of them yes, but I still experience out-of-resources issues.

PR #263, opened 2022-08-15
Add timespan logging via TimespanLogging.jl

jpsamaroo:
This will make it easier to avoid needing to use tools like `rocprof` to do simple timespan analysis of actions like kernel launch, allocation latency, memory transfer time, etc.

Todo:
- [x] Ensure that all starts and finishes can be easily paired
- [x] Benchmark cost of disabled logging vs. master
- [x] Add opt-in preference to statically enable logging at precompile time
- [x] Add docs
- [x] Add tests


jpsamaroo:
The latest revision of this PR now provides a Preferences option to statically enable logging (off by default), and with it disabled, logging does not have any measurable overhead. With logging statically enabled (but not actively logging data somewhere), I see roughly 30% more allocations and 10% more runtime; with logging fully enabled and writing, allocations are 10x greater and runtime is 50% slower. Further work can definitely bring that time down significantly, but I think this is looking good for merging quite soon.

PR #264, opened 2022-08-16
Run wavefront tests with detected wavefrontsize

torrance:
Previously this was hardcarded to a wavefrontsize of 64, however some older cards use 32.

### DIFF
--- a/src/device.jl
+++ b/src/device.jl
@@ -145,8 +145,8 @@ function device_type(agent::HSA.Agent)
     return devtype[]
 end
 
-wavefront_size(device::ROCDevice) = wavefront_size(device.agent)
-function wavefront_size(agent::HSA.Agent)
+wavefrontsize(device::ROCDevice) = wavefrontsize(device.agent)
+function wavefrontsize(agent::HSA.Agent)
     wfsize = Ref{UInt32}()
     getinfo(agent, HSA.AGENT_INFO_WAVEFRONT_SIZE, wfsize) |> check
     return wfsize[]
--- a/src/device.jl
+++ b/src/device.jl
@@ -145,6 +145,13 @@ function device_type(agent::HSA.Agent)
     return devtype[]
 end
 
+wavefront_size(device::ROCDevice) = wavefront_size(device.agent)
+function wavefront_size(agent::HSA.Agent)
+    wfsize = Ref{UInt32}()
+    getinfo(agent, HSA.AGENT_INFO_WAVEFRONT_SIZE, wfsize) |> check
+    return wfsize[]
+end
+
 function getinfo(agent::HSA.Agent, attribute::HSA.AgentInfo,
                  value::Union{Vector,Base.RefValue,String})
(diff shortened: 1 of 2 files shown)


test/device/wavefront.jl
```diff
@@ -1,4 +1,11 @@
 @testset "Wavefront Operations" begin
+    wavefrontsize = Ref{UInt32}()
+    AMDGPU.HSA.LibHSARuntime.hsa_agent_get_info(
+        AMDGPU.default_device().agent,
+        AMDGPU.HSA.AGENT_INFO_WAVEFRONT_SIZE,
+        wavefrontsize
+    )
```
jpsamaroo, line 7: Would you mind adding a method `getinfo(::ROCDevice, ..., ...)` in `src/device.jl` that allows us to pass `AMDGPU.default_device()`? And then this can be:
```julia
wavefrontsize = Ref{UInt32}()
AMDGPU.Runtime.getinfo(AMDGPU.default_device(), AMDGPU.HSA.AGENT_INFO_WAVEFRONT_SIZE, wavefrontsize)
```

We could also add a `wavefrontsize(::ROCDevice)` function which does this for us, which would be really convenient.

jpsamaroo (approved):
Looking good!

torrance:
I've pushed through a common `wavefront_size(device)` function and set the tests to make use of this.

However, I've held off on writing a `getinfo(::Device, ...)` because I think that this function should be written in such a way that mutable reference values do not need to be passed in (i.e. require underlying knowledge of the return type), rather than just be a wrapper for getinfo with agent -> device.

Going forward, either we can continue writing wrappers for getinfo in the style of 'wavefront_size` as they become necessary, or instead write getinfo to know the return types/sizes of each of its possible enumerations. I think the first option is more reasonable so long as we're only querying a minority of these enums.

src/device.jl
```diff
@@ -145,6 +145,13 @@ function device_type(agent::HSA.Agent)
     return devtype[]
 end
 
+wavefront_size(device::ROCDevice) = wavefront_size(device.agent)
+function wavefront_size(agent::HSA.Agent)
+    wfsize = Ref{UInt32}()
+    getinfo(agent, HSA.AGENT_INFO_WAVEFRONT_SIZE, wfsize) |> check
+    return wfsize[]
+end
+
```
jpsamaroo, line 154: Can we call this `wavefrontsize` instead of `wavefront_size`? It should be possible to merge this with `Device.wavefrontsize()` in a future PR, so it'd be great if they had the same name.

jpsamaroo (approved):
Looks great! Just one nitpick.

jpsamaroo:
Also, I'm in agreement that we should be able to call `getinfo` without providing a `Ref` object, and let it determine and create the right allocation instead; filed as https://github.com/JuliaGPU/AMDGPU.jl/issues/273

jpsamaroo:
Thanks!

PR #265, opened 2022-08-17
Ensure killqueue!() updated global QUEUES list

torrance:
Previously, a call to `killqueue()` would not clean up the global `QUEUES` list. On my device, a new call to `Queue()` would reuse the same queue pointer, and this would cause `Queue()` to throw an error since it detected duplicate queue pointers in the `QUEUES` list. This was causing tests related to queues to fail.

This PR ensures `killqueue()` will clean up the `QUEUES` list.

It also adds documentation and cleanup to `kill_queues()`.

### DIFF
--- a/src/queue.jl
+++ b/src/queue.jl
@@ -132,8 +132,6 @@ attempt has already been made. Since queue pointers may be reused by the device,
 this may cause unpredictable behaviour by destroying an active queue.
 """
 function kill_queue!(queue::ROCQueue; force=false)
-    #queue_ptr = @atomicswap queue.queue = Ptr{HSA.Queue}(0)
-    queue_ptr = queue.queue
     if !force && !replacefield!(queue, :active, true, false, :acquire_release, :acquire).success
         # The queue was already marked as inactive and we are not forcing it.
         return false
@@ -151,14 +149,14 @@ function kill_queue!(queue::ROCQueue; force=false)
 
         # Send exception to all waiter signals
         if queue.status != HSA.STATUS_SUCCESS
-            err = QueueError(queue_ptr, HSAError(queue.status))
+            err = QueueError(queue.queue, HSAError(queue.status))
             for signal in _active_kernels[queue]
                 signal.exception = err
                 notify(signal)
             end
         end
     end
-    HSA.queue_destroy(queue_ptr) |> check
+    HSA.queue_destroy(queue.queue) |> check
 
     return true
 end
--- a/src/queue.jl
+++ b/src/queue.jl
@@ -134,6 +134,7 @@ function kill_queue!(queue::ROCQueue; force=false)
         if get(DEFAULT_QUEUES, queue.device, nothing) == queue
             delete!(DEFAULT_QUEUES, queue.device)
         end
+        delete!(QUEUES, queue.queue)
         close(queue.cond)
 
         # Send exception to all waiter signals
--- a/src/queue.jl
+++ b/src/queue.jl
@@ -89,14 +89,6 @@ function ROCQueue(device::ROCDevice; priority::Symbol=:normal)
     AMDGPU.hsaref!()
     finalizer(queue) do queue
         kill_queue!(queue)
-        while !trylock(RT_LOCK)
-        end
-        try
-            delete!(QUEUES, queue_ptr)
-            delete!(_active_kernels, queue)
-        finally
-            unlock(RT_LOCK)
-        end
         AMDGPU.hsaunref!()
     end
     return queue
@@ -139,12 +131,6 @@ function kill_queue!(queue::ROCQueue; force=false)
     @atomic queue.active = false
 
     lock(RT_LOCK) do
-        # Clean up global queue lists
-        if get(DEFAULT_QUEUES, queue.device, nothing) == queue
-            delete!(DEFAULT_QUEUES, queue.device)
-        end
-        delete!(QUEUES, queue.queue)
-
         close(queue.cond)
 
         # Send exception to all waiter signals
@@ -155,6 +141,13 @@ function kill_queue!(queue::ROCQueue; force=false)
                 notify(signal)
             end
         end
+
+        # Clean up global queue lists
+        if get(DEFAULT_QUEUES, queue.device, nothing) == queue
+            delete!(DEFAULT_QUEUES, queue.device)
+        end
+        delete!(QUEUES, queue.queue)
+        delete!(_active_kernels, queue)
     end
     HSA.queue_destroy(queue.queue) |> check
 
--- a/src/queue.jl
+++ b/src/queue.jl
@@ -121,20 +121,32 @@ function get_default_queue(device::ROCDevice)
     end
 end
 
-"Kills all kernels executing on the given queue, and destroys the queue."
+"""
+    kill_queue!(queue::ROCQueue; force=false)
+
+Kills all kernels executing on the given queue, destroys the queue,
+and performs internal clean up. Returns true upon success.
+
+If `force` is `true`, this will attempt to kill a queue even if a previous
+attempt has already been made. Since queue pointers may be reused by the device,
+this may cause unpredictable behaviour by destroying an active queue.
+"""
 function kill_queue!(queue::ROCQueue; force=false)
     #queue_ptr = @atomicswap queue.queue = Ptr{HSA.Queue}(0)
     queue_ptr = queue.queue
     if !force && !replacefield!(queue, :active, true, false, :acquire_release, :acquire).success
-        return false # We didn't destroy the queue
+        # The queue was already marked as inactive and we are not forcing it.
+        return false
     end
     @atomic queue.active = false
 
     lock(RT_LOCK) do
+        # Clean up global queue lists
         if get(DEFAULT_QUEUES, queue.device, nothing) == queue
             delete!(DEFAULT_QUEUES, queue.device)
         end
         delete!(QUEUES, queue.queue)
+
         close(queue.cond)
 
         # Send exception to all waiter signals
@@ -148,5 +160,5 @@ function kill_queue!(queue::ROCQueue; force=false)
     end
     HSA.queue_destroy(queue_ptr) |> check
 
-    return true # We actually destroyed the queue
+    return true
 end
--- a/src/queue.jl
+++ b/src/queue.jl
@@ -57,14 +57,10 @@ function ROCQueue(device::ROCDevice; priority::Symbol=:normal)
             err isa EOFError || rethrow(err)
             return
         end
+
         lock(RT_LOCK) do
-            if !haskey(QUEUES, queue_ptr) || QUEUES[queue_ptr].value === nothing
-                # Queue was destroyed
-                return
-            end
-            queue = QUEUES[queue_ptr].value
-            if queue.status != HSA.STATUS_SUCCESS
-                # Queue encountered an async error, manually kill it
+            if (@atomic queue.active) && queue.status != HSA.STATUS_SUCCESS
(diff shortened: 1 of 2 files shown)


src/queue.jl
```diff
@@ -121,31 +121,42 @@ function get_default_queue(device::ROCDevice)
...
+        return false
     end
     @atomic queue.active = false
 
     lock(RT_LOCK) do
+        # Clean up global queue lists
         if get(DEFAULT_QUEUES, queue.device, nothing) == queue
             delete!(DEFAULT_QUEUES, queue.device)
         end
+        delete!(QUEUES, queue.queue)
```
jpsamaroo, line 146: We already do this in the `finalizer` call above in the `ROCQueue` constructor; if we do this in `kill_queue!`, then multiple tasks might try to delete it, and some of them will fail with an error. My intention for putting it in the finalizer is to ensure that only one task does this deletion, which only happens once there are no longer any references to that queue.

test/hsa/queue.jl
```diff
@@ -7,4 +7,13 @@
         end
         @test_throws ArgumentError ROCQueue(device; priority=:fake)
     end
+
+    @testset "KillQueue cleanup" begin
```
jpsamaroo, line 11:
```suggestion
    @testset "kill_queue! cleanup" begin
```

jpsamaroo (changes requested):
I'm not sure that this is correct, although I do agree that something is broken.

src/queue.jl
jpsamaroo: After thinking through this, I understand now why we're getting this assertion in CI. Because we call `HSA.queue_destroy` within `kill_queue!`, the HSA runtime has the right to re-use that same address to serve future queue allocations. When we allocate queues in a tight loop, that can cause a queue to be allocated with the same address as another queue which was just destroyed (but not yet deleted from `QUEUES`).

I think we should either:
- Destroy the queue in the finalizer, but set `queue.queue = C_NULL` in `kill_queue!` to signal that the queue is effectively dead (probably not preferred because now `kill_queue!` doesn't stop a runaway kernel, which is why I wrote it)
- Ensure we always check if `QUEUES` has our queue pointer before trying to access/delete `QUEUES`, and then we can do this in `kill_queue!` again like you have done here.

torrance: I was seeing the errors not because of a tight loop, but because during testing the "HSA Async Queue Error" was deliberately triggering a QueueError and as part of its error handler it was calling `kill_queue!`. However, it looks like the `Queue` object has been GC'ed by the time `test/hsa/queue.jl` ran (and so it's finalizer hadn't run either), and so then I was seeing the reuse of the pointer during these tests.

As a third option to your suggestions, I suggest simply moving all the finalizer code to `kill_queue()`. This is OK since this function already protects against being run multiple times by its atomic check on its `active` state. (And additionally, calling `delete()` on a dictionary does not raise any errors if the key doesn't exist).

src/queue.jl
```diff
@@ -121,31 +121,42 @@ function get_default_queue(device::ROCDevice)
...
+Kills all kernels executing on the given queue, destroys the queue,
+and performs internal clean up. Returns true upon success.
+
+If `force` is `true`, this will attempt to kill a queue even if a previous
+attempt has already been made. Since queue pointers may be reused by the device,
+this may cause unpredictable behaviour by destroying an active queue.
+"""
 function kill_queue!(queue::ROCQueue; force=false)
-    #queue_ptr = @atomicswap queue.queue = Ptr{HSA.Queue}(0)
-    queue_ptr = queue.queue
```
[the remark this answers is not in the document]
torrance: I don't fully understand the workflow you have in mind here: I removed it because I assumed the queue pointer wouldn't change.

Isn't allowing this pointer to change elsewhere asking for trouble? If it's set to NULL at some point during `kill_queue()`, does that mean our cached reference to it is still okay? Could we now inadvertently be killing a new queue with a reused pointer? Seems like we're making life hard from the perspective of thread safety. (For that matter, I also feel like the 'force' option here is asking for trouble).

torrance:
Please see latest commits. I've centralised all `ROCQueue` cleanup in `kill_queue!()`. However, I'm concerned about your comment regarding the queue pointer being nulled during this function, and I don't know how to handle that without knowing more.

jpsamaroo:
Ok, here's all the mechanisms in play that I think we need to consider:
- Kernel launch needs to query and access the queue, but can yield indefinitely due to the queue becoming full, so it's possible that the queue has entered an error (not yet dead) state by the time we yield back. We therefore need a fast check for queue validity before querying the queue (which we have), and one just before we write our packet to the queue (which we do not yet have). Also, we implicitly preserve the `ROCQueue` object during launch, so we shouldn't ever see the finalizer concurrently run with kernel launches (and more importantly, the queue pointer should not be re-used during that time).
- The queue error monitor (with HSA calling `queue_error_handler` and Julia subsequently returning from https://github.com/JuliaGPU/AMDGPU.jl/blob/287712e571cbac0c58db3debe702cff50ed95f6e/src/queue.jl#L55) can be triggered at any time, and does not indicate that the queue is destroyed; that only happens once `HSA.queue_destroy` is called. N.B. We cannot take locks in `queue_error_handler`, as it runs on another thread.
- The queue error monitor queries `QUEUES` as a means to prevent the task from inadvertently preserving the `ROCQueue` object beyond its useful lifetime. Therefore, the lifetime of `QUEUES` entries must be explicitly managed to correlate with the lifetime of the associated `ROCQueue` object; the most obvious way to do this is to only allow the finalizer to delete the slot for its queue.

jpsamaroo:
Also, very importantly:
- `HSA.queue_destroy` means that the queue is truly dead and its pointer can be re-used, so we better not have that pointer be accessed anywhere by the time that call happens.

pxl-th:
One situation that causes this issue and is present in the tests as well is the following:

1. We are [launching](https://github.com/JuliaGPU/AMDGPU.jl/blob/166043c4e689c330f275fb7419f65471677b8929/test/hsa/error.jl#L8) kernel and it also creates default queue which is added both to `QUEUES` & `DEFAULT_QUEUES`.
2. While waiting on it, this kernel timeouts, [sending](https://github.com/JuliaGPU/AMDGPU.jl/blob/166043c4e689c330f275fb7419f65471677b8929/src/kernel-signal.jl#L53) `SignalTimeoutException`. We kill the respective queue, which removes it from `DEFAULT_QUEUES`, but not `QUEUES`.
3. Then another test [creates](https://github.com/JuliaGPU/AMDGPU.jl/blob/166043c4e689c330f275fb7419f65471677b8929/test/hsa/queue.jl#L6) several queues in a row, failing on the first, because `HSA.create_queue` created queue with the same pointer as in `QUEUES`. We fail.
4. Only sometime after the finalizer is called for the queue created at 1. step, removing it from `QUEUES`.

So we kill queue, but it is not removed from `QUEUES` until finalizer is called.
Is there a situation where we might need killed queue to remain in `QUEUES`?
If not, then it is probably safe to remove it from `QUEUES` as we kill it.

pxl-th:
One situation that causes this issue and is present in the tests as well is the following:

1. We are [launching](https://github.com/JuliaGPU/AMDGPU.jl/blob/166043c4e689c330f275fb7419f65471677b8929/test/hsa/error.jl#L8) kernel and it also creates default queue which is added both to `QUEUES` & `DEFAULT_QUEUES`.
2. While waiting on it, this kernel timeouts, [sending](https://github.com/JuliaGPU/AMDGPU.jl/blob/166043c4e689c330f275fb7419f65471677b8929/src/kernel-signal.jl#L53) `SignalTimeoutException`. We kill the respective queue, which removes it from `DEFAULT_QUEUES`, but not `QUEUES`.
3. Then another test [creates](https://github.com/JuliaGPU/AMDGPU.jl/blob/166043c4e689c330f275fb7419f65471677b8929/test/hsa/queue.jl#L6) several queues in a row, failing on the first, because `HSA.create_queue` created queue with the same pointer as in `QUEUES`. We fail.
4. Only sometime after the finalizer is called for the queue created at 1. step, removing it from `QUEUES`.

So we kill queue, but it is not removed from `QUEUES` until finalizer is called.
Is there a situation where we might need killed queue to remain in `QUEUES`?
If not, then it is probably safe to remove it from `QUEUES` as we kill it.

Here's example from the logs (hard kill indicating that `HSA.queue_destroy` was called)

```julia
[ Info: [queue] Getting default queue
[ Info: [queue] creation
[ Info: [queue] Getting default queue
[ Info: [queue] killing queue, force=false, active=true
[ Info: [queue] hard kill
[ Info: [queue] creation
Priorities: Error During Test at /home/pxl-th/.julia/dev/AMDGPU/test/hsa/queue.jl:2
  Got exception outside of a @test
  AssertionError: !(haskey(QUEUES, queue.queue))
  Stacktrace:
    [1] (::AMDGPU.Runtime.var"#19#23")()
      @ AMDGPU.Runtime ~/.julia/dev/AMDGPU/src/queue.jl:51
    [2] lock(f::AMDGPU.Runtime.var"#19#23", l::ReentrantLock)
      @ Base ./lock.jl:229
    [3] ROCQueue(device::ROCDevice; priority::Symbol)
      @ AMDGPU.Runtime ~/.julia/dev/AMDGPU/src/queue.jl:50
    [4] macro expansion
      @ ~/.julia/dev/AMDGPU/test/hsa/queue.jl:6 [inlined]
    [5] macro expansion
      @ ~/bin/julia-latest/share/julia/stdlib/v1.9/Test/src/Test.jl:1493 [inlined]
    [6] macro expansion
      @ ~/.julia/dev/AMDGPU/test/hsa/queue.jl:3 [inlined]
    [7] macro expansion
      @ ~/bin/julia-latest/share/julia/stdlib/v1.9/Test/src/Test.jl:1493 [inlined]
    [8] top-level scope
      @ ~/.julia/dev/AMDGPU/test/hsa/queue.jl:2
    [9] include(fname::String)
      @ Base.MainInclude ./client.jl:478
   [10] macro expansion
      @ ~/.julia/dev/AMDGPU/test/runtests.jl:61 [inlined]
   [11] macro expansion
      @ ~/bin/julia-latest/share/julia/stdlib/v1.9/Test/src/Test.jl:1493 [inlined]
   [12] macro expansion
      @ ~/.julia/dev/AMDGPU/test/runtests.jl:58 [inlined]
   [13] macro expansion
      @ ~/bin/julia-latest/share/julia/stdlib/v1.9/Test/src/Test.jl:1493 [inlined]
   [14] top-level scope
      @ ~/.julia/dev/AMDGPU/test/runtests.jl:54
   [15] include(fname::String)
      @ Base.MainInclude ./client.jl:478
   [16] top-level scope
      @ none:6
   [17] eval
      @ ./boot.jl:370 [inlined]
   [18] exec_options(opts::Base.JLOptions)
      @ Base ./client.jl:280
   [19] _start()
      @ Base ./client.jl:522
Test Summary:             | Pass  Error  Total   Time
AMDGPU                    |    2      1      3  19.5s
  HSA                     |    2      1      3  19.5s
    HSA Status Error      |    1             1   0.0s
    HSA Async Queue Error |    1             1  18.4s
    Queues                |           1      1   0.7s
      Priorities          |           1      1   0.7s
  Device Functions        |               None   0.0s
ERROR: LoadError: Some tests did not pass: 2 passed, 0 failed, 1 errored, 0 broken.
in expression starting at /home/pxl-th/.julia/dev/AMDGPU/test/runtests.jl:49
[ Info: [queue] Finalizer
```

jpsamaroo:
Closing as this seems to be fixed in #306; thanks for looking into this @torrance !

PR #266, opened 2022-08-17
During HostCall, ensure device has finished using buffers before freeing

torrance:
As was warned in the comments in `device/gc/hostcall.jl`, these `free()` operations were indeed dangerous: I was seeing irregular but routine segfualts when running the tests that were a result of freeing too early.

This `HostCall()` code rather confusing to me, but I think (??) I've done the right thing here by requiring the signal to shift from `HOST_MSG_SENTINEL` to either an OK or ERROR code by the device. At least, I can no longer reproduce the segfaults!

I've also removed the calls to `sleep(1)` in the tests, which I assume were a mitigation against these race conditions.

### DIFF
--- a/src/device/gcn/hostcall.jl
+++ b/src/device/gcn/hostcall.jl
@@ -327,15 +327,19 @@ function HostCall(func, rettype, argtypes; return_task=false,
             HSA.signal_store_screlease(signal.signal[], HOST_ERR_SENTINEL)
             rethrow(err)
         finally
-            # TODO: This is probably a bad idea to free while the kernel might
-            # still hold a reference. We should probably have some way to
-            # reference count these buffers from kernels (other than just
-            # using a ROCArray, which can't currently be serialized to the
-            # GPU).
-            if isassigned(ret_buf)
-                Mem.free(ret_buf[])
+            # We need to free the memory buffers, but first we need to ensure that
+            # the device has read from these buffers. Therefore we wait either for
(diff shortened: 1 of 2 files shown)


src/device/gcn/hostcall.jl
```diff
@@ -327,15 +327,19 @@ function HostCall(func, rettype, argtypes; return_task=false,
...
-            if isassigned(ret_buf)
-                Mem.free(ret_buf[])
+            # We need to free the memory buffers, but first we need to ensure that
+            # the device has read from these buffers. Therefore we wait either for
+            # READY_SENTINEL or else an error signal.
+            while !Runtime.RT_EXITING[]
+                prev = HSA.signal_load_scacquire(signal.signal[])
+
+                if prev == READY_SENTINEL || prev == HOST_ERR_SENTINEL || prev == DEVICE_ERR_SENTINEL
+                    Mem.free(ret_buf[])
```
jpsamaroo, line 337:
```suggestion
                    if isassigned(ret_buf)
                        Mem.free(ret_buf[])
                    end
```

We need to make sure that we handle the situation where we hit an error before assigning `ret_buf`.

jpsamaroo (approved):
This is an awesome catch!

torrance:
I've commit your check for `isassigned()`. :)

jpsamaroo:
Thanks!

PR #268, opened 2022-08-18
Remove code that duplicates AbstractFFTs; add tests for casting

torrance:
This code is duplicated from AbstractFFTs; no need to copy it. (Additionally, `Float16` was misspelled as `FLoat16`.)

Add tests for casting to complex floats for FFTs, and for handing Integer inputs.

Note: Float16 is not tested as the casting is broken upstream in AbstractFFTs.

### DIFF
--- a/test/rocarray/fft.jl
+++ b/test/rocarray/fft.jl
@@ -246,7 +246,7 @@ end
 
 end
 
-@testset "Casted types" begin
+@testset "Promoted types" begin
     @testset for T in (Float32, Float64)
         X = rand(T, 10, 10)
         d_X = ROCArray(X)
--- a/test/rocarray/fft.jl
+++ b/test/rocarray/fft.jl
@@ -245,4 +245,42 @@ end
 end
 
 end
+
+@testset "Casted types" begin
+    @testset for T in (Float32, Float64)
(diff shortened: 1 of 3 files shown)


test/rocarray/fft.jl
```diff
@@ -245,4 +245,42 @@ end
 end
 
 end
+
+@testset "Casted types" begin
```
jpsamaroo, line 249: In what sense are these "Casted types"? It seems like we're not casting these to another type, although I'm not super familiar with the FFT code.

jpsamaroo (approved):
Awesome!

jpsamaroo:
I'm going to push some changes to master that will let us get CI properly testing FFTs (since right now we only have rocFFT testing enabled on 1.8 and nightly/1.9).

test/rocarray/fft.jl
torrance: You're right, they're not casted, they're promoted. E.g. Int -> Complex64, Rational -> ComplexF64, etc.  (see: https://github.com/JuliaMath/AbstractFFTs.jl/blob/master/src/definitions.jl#L205)

I'll update the PR to reflect this language.

jpsamaroo:
Thanks!

PR #282, opened 2022-09-11
Use an optimized norm function for ROCBLASArray

amontoison:


src/blas/highlevel.jl
```diff
@@ -43,7 +43,7 @@ function LinearAlgebra.BLAS.dotu(DX::ROCArray{T}, DY::ROCArray{T}) where T<:Unio
     dotu(n, DX, 1, DY, 1)
 end
 
-# FIXME: LinearAlgebra.norm(x::ROCBLASArray) = nrm2(x)
+LinearAlgebra.norm(x::ROCArray{T}) where T<:ROCBLASFloat = nrm2(length(x), x, 1)
```
jpsamaroo, line 46:
```suggestion
LinearAlgebra.norm(x::ROCArray{T}) where T<:Union{ROCBLASFloat,ROCBLASComplex} = nrm2(length(x), x, 1)
```

jpsamaroo (approved):
Looks great! Just missed making the `norm` implementation support Complex numbers.

src/blas/highlevel.jl
jpsamaroo: Note that this works because there's probably a `norm` implementation in GPUArrays, but that would compile a kernel, which isn't necessary.

jpsamaroo:
Thanks!

src/blas/highlevel.jl
amontoison: @jpsamaroo 
Is it normal that `ROCBLASFloat = Union{Float32,Float64}`?
https://github.com/JuliaGPU/AMDGPU.jl/blob/master/src/blas/librocblas_types.jl#L16

In general we have:
```julia
BlasFloat = Union{Float32, Float64, ComplexF32, ComplexF64}
BlasReal = Union{Float32, Float64}
BlasComplex = Union{ComplexF32, ComplexF64}
```

jpsamaroo: That's a good point; want to send in a PR for this? It'll be a lot of automatic search-and-replace.

PR #284, opened 2022-09-12
Add rocBLAS_jll and rocSPARSE_jll deps

jpsamaroo:
And upper-bound all ROCm JLLs to 5 in preparation for upcoming ROCm 5.x JLLs

@pxl-th


deps/build.jl
```diff
@@ -378,39 +384,31 @@ function main()
 
         if !use_artifacts
```
pxl-th, line 385: I think this is not needed. As it will never hit `use_artufacts` case below.

jpsamaroo: Thanks for catching this!

deps/build.jl
```diff
@@ -376,41 +382,31 @@ function main()
...
         config[:hip_configured] = true
 
-        if !use_artifacts
-            ### Find external HIP-based libraries
-            for name in ("rocblas", "rocsolver", "rocsparse", "rocalution", "rocfft", "MIOpen")
-                lib = Symbol("lib$(lowercase(name))")
-                path = find_rocm_library("lib$name", roc_dirs)
+        ### Find external HIP-based libraries
+        for (name, pkg) in rocm_ext_libs
+            lib = Symbol("lib$(lowercase(name))")
```
pxl-th, line 387:
```suggestion
            llib = Symbol(:lib, lowercase(string(name)))
```

deps/build.jl
```diff
@@ -376,41 +382,31 @@ function main()
...
+            if use_artifacts && pkg !== nothing
+                try
+                    @eval using $pkg
+                    config[lib] = Symbol(pkg, :., lib)
```
pxl-th, line 391:
```suggestion
                    config[lib] = getfield(@eval($pkg), lib)
```

PR #289, opened 2022-09-15
CI: Add gfx90a to more jobs

jpsamaroo:
Since we're having trouble with the gfx908 runners, let's see if our gfx90a agents fare any better.


jpsamaroo:
The issues with gfx90a should be fixed with HIP_jll 5.2.3, and the ROCR assertions seem to also be fixed with 5.2.3.

PR #296, opened 2022-09-24
Do not rethrow EOF

pxl-th:
- Do not rethrow `EOFError`.
This happens when Julia process is exiting, but hostcalls are sleeping.

This gets rid of confusing error messages at the end of program execution:
```
┌ Error: HostCall error
│   exception =
│    EOFError: read end of file
│    Stacktrace:
│     [1] wait
│       @ ./asyncevent.jl:155 [inlined]
│     [2] sleep(sec::Float64)
│       @ Base ./asyncevent.jl:240
│     [3] hostcall_host_wait(signal::ROCSignal; maxlat::Float64, timeout::Nothing)
│       @ AMDGPU.Device ~/.julia/dev/AMDGPU/src/device/gcn/hostcall.jl:441
│     [4] macro expansion
│       @ ~/.julia/dev/AMDGPU/src/device/gcn/hostcall.jl:325 [inlined]
│     [5] (::AMDGPU.Device.var"#90#91"{ROCDevice, Float64, Nothing, Bool, AMDGPU.Compiler.var"#43#47", DataType, DataType, AMDGPU.Device.HostCall{UInt64, Int64, Tuple{Core.LLVMPtr{UInt8, 1}}}, ROCSignal})()
│       @ AMDGPU.Device ./threadingconstructs.jl:322
└ @ AMDGPU.Device ~/.julia/dev/AMDGPU/src/device/gcn/hostcall.jl:378
```

- Change `QueueError` to `SignalTimeoutException` in HSA error test. @jpsamaroo, without timeout that kernel never ends. So I think this is the error to check for.

### DIFF
--- a/test/hsa/error.jl
+++ b/test/hsa/error.jl
@@ -5,5 +5,5 @@ end
 
 @testset "HSA Async Queue Error" begin
     kernel() = (Device.trap(); nothing)
-    @test_throws Runtime.SignalTimeoutException wait(@roc kernel())
+    @test_throws Runtime.QueueError wait(@roc kernel())
 end
--- a/test/hsa/error.jl
+++ b/test/hsa/error.jl
@@ -5,5 +5,5 @@ end
 
 @testset "HSA Async Queue Error" begin
     kernel() = (Device.trap(); nothing)
-    @test_throws Runtime.QueueError wait(@roc kernel())
+    @test_throws Runtime.SignalTimeoutException wait(@roc kernel())
 end
--- a/src/device/gcn/hostcall.jl
+++ b/src/device/gcn/hostcall.jl
@@ -375,9 +375,13 @@ function HostCall(func, rettype, argtypes; return_task=false,
             end
         catch err
             # Gracefully terminate all waiters
-            @error "HostCall error" exception=(err,catch_backtrace())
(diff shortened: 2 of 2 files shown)


test/hsa/error.jl
```diff
@@ -5,5 +5,5 @@ end
 
 @testset "HSA Async Queue Error" begin
     kernel() = (Device.trap(); nothing)
-    @test_throws Runtime.QueueError wait(@roc kernel())
+    @test_throws Runtime.SignalTimeoutException wait(@roc kernel())
```
jpsamaroo, line 8: A trap should trigger the async queue handler, which will then throw a `QueueError` at all running kernels (https://github.com/JuliaGPU/AMDGPU.jl/blob/c85bc0a2ef0b9a939016a71e201c0ded0d1f58c1/src/queue.jl#L140-L146). If this isn't happening, then it's a bug.

jpsamaroo (approved):
Looks great! Just one concern about the change to the "HSA Async Queue Error" test.

test/hsa/error.jl
pxl-th: OK, then it is a bug. Reverted this test.

jpsamaroo:
Thanks!

PR #297, opened 2022-09-25
Use correct queue for kernels

pxl-th:
In [`macro roc`](https://github.com/JuliaGPU/AMDGPU.jl/blob/c85bc0a2ef0b9a939016a71e201c0ded0d1f58c1/src/highlevel.jl#L313) `signal_kwargs` does not contain `queue` argument, which leads to usage of the default queue in [`create_event`](https://github.com/JuliaGPU/AMDGPU.jl/blob/c85bc0a2ef0b9a939016a71e201c0ded0d1f58c1/src/highlevel.jl#L106) thus making `queue` args basically useless.

To fix this I added `:queue` to `signal_kwargs` and use signal's queue when calling kernels.
`:queue` in `compiler_kwargs` and `call_kwargs` is thus not needed.

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -67,7 +67,7 @@ end
 
 @inline function roccall(kernel::HostKernel, tt, args...; config=nothing, signal, device=nothing, kwargs...)
     device = something(device, AMDGPU.default_device())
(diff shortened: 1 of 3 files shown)


jpsamaroo:
Thanks!

PR #300, opened 2022-09-27
Simplify kernel dispatch

pxl-th:
- Simplify kernel dispatch logic.
`ROCKernelSignal` contains all the necessary information for dispatching kernels (like device and queue).
No need for repeating 'device' & 'queue' arguments.
This reduces points of failure as well.

- Create queue on the correct device.
Previously `create_event` function would create queue without taking `device` into account.

- Check that passed device and queue are valid.
If passing both `device` and `queue`, ensure that `queue` is on the `device`, otherwise raise an error.
So something like below, will raise an error:
```julia
non_default_device = ... # non-default device
default_queue = AMDGPU.default_queue()
@roc queue=default_queue device=non_default_device f()
```

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -65,13 +65,14 @@ end
 
 @doc (@doc AbstractKernel) HostKernel
 
-@inline function roccall(kernel::HostKernel, tt, args...; config=nothing, signal, device=nothing, kwargs...)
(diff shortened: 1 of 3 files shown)


pxl-th:
I will do this in a different PR.

PR #294, opened 2022-09-22
Fix precompilation

pxl-th:
On Julia 1.9 when using AMDGPU & ROCKernels in another project, ROCKernels fails to precompile (see error below). Specifically, it complains at `src/AMDGPU.jl:231`.

We can use `hsa_version()` function directly instead of storing it in the variable during `__init__`. I didn't see `libhsaruntime_version` being used anywhere else.

```
Precompiling project...
  ✗ ROCKernels
  ✗ INGP
  1 dependency successfully precompiled in 14 seconds. 184 already precompiled.

ERROR: The following 2 direct dependencies failed to precompile:

INGP [3003bbbd-5e0c-4496-81c4-ba6063c314ae]

Failed to precompile INGP [3003bbbd-5e0c-4496-81c4-ba6063c314ae] to /home/pxl-th/.julia/compiled/v1.9/INGP/jl_Lzl3vl.
ERROR: LoadError: InitError: Evaluation into the closed module `AMDGPU` breaks incremental compilation because the side effects will not be permanent. This is likely due to some other module mutating `AMDGPU` with `eval` during precompilation - don't do this.
Stacktrace:
  [1] eval
    @ ./boot.jl:370 [inlined]
  [2] __init__()
    @ AMDGPU ~/.julia/dev/AMDGPU/src/AMDGPU.jl:231
  [3] _include_from_serialized(pkg::Base.PkgId, path::String, depmods::Vector{Any})
    @ Base ./loading.jl:928
  [4] _require_search_from_serialized(pkg::Base.PkgId, sourcepath::String, build_id::UInt64)
    @ Base ./loading.jl:1136
  [5] _require(pkg::Base.PkgId, env::String)
    @ Base ./loading.jl:1412
  [6] _require_prelocked(uuidkey::Base.PkgId, env::String)
    @ Base ./loading.jl:1298
  [7] macro expansion
    @ ./loading.jl:1278 [inlined]
  [8] macro expansion
    @ ./lock.jl:267 [inlined]
  [9] require(into::Module, mod::Symbol)
    @ Base ./loading.jl:1241
 [10] top-level scope
    @ ~/code/INGP.jl/src/INGP.jl:51
 [11] include
    @ ./Base.jl:439 [inlined]
 [12] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt64}}, source::Nothing)
[...]

PR #306, opened 2022-10-06
Remove waiter and error monitor threads

### LINKED ISSUE #299: Repetetive `AMDGPU.ones` calls crash runtime
Following code snippet will crash:

```julia
julia> using AMDGPU

julia> for _ in 1:100_000
           AMDGPU.ones(Float32, 16);
       end
```

with this error:

```julia
ERROR: schedule: Task not runnable
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] schedule(t::Task, arg::Any; error::Bool)
    @ Base ./task.jl:829
  [3] schedule
    @ ./task.jl:827 [inlined]
  [4] uv_writecb_task(req::Ptr{Nothing}, status::Int32)
    @ Base ./stream.jl:1166
  [5] poptask(W::Base.IntrusiveLinkedListSynchronized{Task})
    @ Base ./task.jl:963
  [6] wait()
    @ Base ./task.jl:972
  [7] wait(c::Base.GenericCondition{ReentrantLock})
    @ Base ./condition.jl:124
  [8] wait(e::Base.Event)
    @ Base ./lock.jl:457
  [9] wait(signal::AMDGPU.Runtime.ROCKernelSignal)
    @ AMDGPU.Runtime ~/.julia/dev/AMDGPU/src/runtime/kernel-signal.jl:51
 [10] gpu_call(::AMDGPU.ROCArrayBackend, f::Function, args::Tuple{ROCVector{Float32}, Float32}, threads::Int64, blocks::Int64; name::Nothing)
    @ AMDGPU ~/.julia/dev/AMDGPU/src/array.jl:16
 [11] #gpu_call#1
    @ ~/.julia/packages/GPUArrays/fqD8z/src/device/execution.jl:65 [inlined]
 [12] gpu_call
    @ ~/.julia/packages/GPUArrays/fqD8z/src/device/execution.jl:34 [inlined]
 [13] fill!
    @ ~/.julia/packages/GPUArrays/fqD8z/src/host/construction.jl:14 [inlined]
 [14] ones
    @ ~/.julia/dev/AMDGPU/src/array.jl:405 [inlined]
 [15] top-level scope
    @ ./REPL[2]:2
```

Putting logs on loots of stuff I noticed that (at least) `ROCSignal` associated with respective `ROCKernelSignals` are not killed. Their finalizers do not run.
I suspect that there may be some dangling references that prevent finalizers from being called.

One thing to note, that the same code, but for CUDA runs fine on a much weaker GPU.

Only when exiting Julia REPL they are triggered. Manually calling `GC.gc()` also does not help.

```
julia> 
[ Info: [main] RT exiting true
[ Info: [kersig] kill ROCSignal(AMDGPU.HSA.LibHSARuntime.hsa_signal_s(0x00007f8d3ded4000))
[ Info: [signal] kill
[ Info: [kersig] kill ROCSignal(AMDGPU.HSA.LibHSARuntime.hsa_signal_s(0x00007f8d3ded8000))
[ Info: [signal] kill
[ Info: [kersig] kill ROCSignal(AMDGPU.HSA.LibHSARuntime.hsa_signal_s(0x00007f8d3deda000))
[ Info: [signal] kill
[ Info: [kersig] kill ROCSignal(AMDGPU.HSA.LibHSARuntime.hsa_signal_s(0x00007f8d3dee4000))
[ Info: [signal] kill
[ Info: [kersig] kill ROCSignal(AMDGPU.HSA.LibHSARuntime.hsa_signal_s(0x00007f8d3def0000))
[ Info: [signal] kill
[ Info: [kersig] kill ROCSignal(AMDGPU.HSA.LibHSARuntime.hsa_signal_s(0x00007f8d3def6000))
[ Info: [signal] kill
[ Info: [kersig] kill ROCSignal(AMDGPU.HSA.LibHSARuntime.hsa_signal_s(0x00007f8dbc7af000))
[ Info: [signal] kill
[ Info: [kersig] kill ROCSignal(AMDGPU.HSA.LibHSARuntime.hsa_signal_s(0x00007f8dcd6fa000))
[ Info: [signal] kill
[ Info: [queue] kill
```

pxl-th:
**Functional changes:**
- Remove error monitor thread from queue.
- Remove waiter thread from `ROCKernelSignal`.
`Base.wait` method now waits directly while still yielding to Julia's scheduler.

These two changes above result in a much more stable runtime and fix #299.
Because they ensure that once the kernel signal is no longer preserved (no other variable holds a reference to it), it can be garbage collected.
It also reduces huge CPU load and random crashes that I previously had.
And I'm able to run all tests on my machine without issues (although some are still failing, but I'm looking into it).

One consequence of this, is that you now have to call `Base.wait` at some point on the signal to release it from global holders, perform cleanup and get exceptions.

It also speeds up some things (below are timings for the first call):

**Before:**
```julia
julia> @time sum(x)
 16.183974 seconds (26.95 M allocations: 1.753 GiB, 3.35% gc time, 34.87% compilation time)
```
**This PR:**
```julia
julia> @time sum(x);
 13.235560 seconds (22.36 M allocations: 1.461 GiB, 3.61% gc time, 39.13% compilation time)
```

- Do not specialize `ROCExecutable` on `Mem.Buffer`.
- Do not specialize `HostCall` on type of signal handle. It is always `UInt64`.
- Minor type instability fixes, like: `Val(x) -> Val{x}()`.
- Simplify kernel dispatch.
`ROCKernelSignal` contains all the necessary information for kernel dispatch. No need for repeated `queue`, `device` arguments.
- `create_event` now correctly takes into account passed `signal`, `queue`, `device` arguments.
Previously it would always create queue on the default device.
- Delete metadata for `ROCModule` in finalizer.
- Output tests now use correct wavefront size instead of assumed 64.

**Refactoring:**
- Move runtime files to `runtime/` directory.
- Move low-level hostcall signal functions to `hostcall_signal_helpers.jl` file.
- Remove unused arguments from some of the functions.
- Remove unused argument from `trap` test kernel.


jpsamaroo (commented):
Some preliminary review comments; more to come once I find time to review the queue/signal/waiter logic changes.

src/AMDGPU.jl
```diff
@@ -281,21 +277,18 @@ function __init__()
...
-                                       "DNN/convolution"))
+    descriptions = (
+        "dense BLAS", "sparse BLAS", "linear solver",
+        "fancy linear solver", "RNG", "FFT", "DNN/convolution")
+    for ((name, pkg), purpose) in zip(rocm_ext_libs, descriptions)
         if use_artifacts && pkg !== nothing && !functional(name)
             # These are numerous and thus noisy
-            build_reason = Symbol(name, :_build_reason)
+            # FIXME crashes on Julia 1.9 due to @eval in precompile
+            # build_reason = Symbol(name, :_build_reason)
```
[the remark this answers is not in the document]
pxl-th: Same issue as here: https://github.com/JuliaGPU/AMDGPU.jl/pull/294

src/array.jl
```diff
@@ -308,7 +306,7 @@ function Base.reshape(a::ROCArray{T,M}, dims::NTuple{N,Int}) where {T,N,M}
 
     Mem.retain(a.buf)
     b = ROCArray{T,N}(a.buf, dims, offset=a.offset, own=false)
-    finalizer(unsafe_free!, b)
+    finalizer(unsafe_free!, b) # TODO is finalizer needed?
```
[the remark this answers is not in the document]
pxl-th: I see, didn't take `Mem.retain` into consideration.

src/runtime/signal.jl
```diff
@@ -0,0 +1,85 @@
...
+    has_timeout = sign(timeout) == 1
+    min_latency = 1_000 # 1 micro-second.
+
+    start_time = time_ns()
+    finished = false
+
+    while !finished
+        v = HSA.signal_wait_scacquire(
+            signal.signal[], HSA.SIGNAL_CONDITION_LT, 1,
+            min_latency, HSA.WAIT_STATE_BLOCKED)
```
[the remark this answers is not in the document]
pxl-th: Not 100% sure, but if the whole process is blocked, would `HSA.WAIT_STATE_BLOCKED` make sense?
https://github.com/RadeonOpenCompute/ROCR-Runtime/blob/adae6c61e10d371f7cbc3d0e94ae2c070cab18a4/src/inc/hsa.h#L1967

src/runtime/signal.jl
```diff
@@ -0,0 +1,85 @@
+const DEFAULT_SIGNAL_TIMEOUT = Ref{Float64}(-1)
```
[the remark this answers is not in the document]
pxl-th: > is there a reason that Union{Float64,Nothing} was removed?

Not really, reverted to `Union{Float64, Nothing}`.

src/device/gcn/hostcall.jl
```diff
@@ -25,114 +25,58 @@ const DEVICE_ERR_SENTINEL = 5
 "Fatal error on host thread accessing the signal."
 const HOST_ERR_SENTINEL = 6
 
-## Hostcall implementation
+const ATOMIC_MONOTONIC = Int32(1)
+const ATOMIC_ACQUIRE = Int32(2)
+const ATOMIC_RELEASE = Int32(3)
+const ATOMIC_ACQ_REL = Int32(4)
+const ATOMIC_SEQ_CST = Int32(5)
```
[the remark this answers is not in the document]
pxl-th: `atomics.jl` is included latest (after `hostcall.jl`) and these variables are only used in hostcall signals.
For now I've moved them to `hostcall_signal_helpers.jl` as they are used only there.

src/device/gcn/hostcall_signal_helpers.jl
```diff
@@ -0,0 +1,82 @@
...
+    signal_handle::UInt64, expected::Int64, value::Int64,
+    order::Int32 = ATOMIC_ACQ_REL,
+)
+    ccall(
+        "extern __ockl_hsa_signal_cas", llvmcall,
+        Int64, (UInt64, Int64, Int64, Int32),
+        signal_handle, expected, value, order)
+end
+
+@inline function store_release!(signal::HSA.Signal, value)
```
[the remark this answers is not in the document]
pxl-th: Agree, renamed these functions using this convention.

src/AMDGPU.jl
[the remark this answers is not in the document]
jpsamaroo: Ok, we should be able to fix this with `getfield(AMDGPU, Symbol(name, :_build_reason))`

test/runtests.jl
```diff
@@ -45,81 +45,81 @@ AMDGPU.versioninfo()
 
 @info "Running tests with $(length(ws)) workers"
 
-push!(tests, "Core" => ()->include("pointer.jl"))
```
jpsamaroo, line 48: Reminder to re-enable these before merge

Manifest.toml.gpuarrays-6
```diff
@@ -1,533 +0,0 @@
-# This file is machine-generated - editing it directly is not advised
```
jpsamaroo, line 1: If we're going to delete these (which I'm fine with), we should remove the corresponding jobs in buildkite CI.

src/device/gcn/hostcall.jl
```diff
@@ -370,9 +364,9 @@ function hostcall_host_wait(signal_handle::HSA.Signal; maxlat=DEFAULT_HOSTCALL_L
     @debug "Hostcall: Waiting on signal $signal_handle"
     start_time = time_ns()
     while !Runtime.RT_EXITING[]
-        prev = get_value(signal_handle)
+        prev = host_signal_load(signal_handle)
         if prev == DEVICE_MSG_SENTINEL
-            prev = set_value_acq_rel!(
+            prev = cmpxchg_acq_rel!(
```
jpsamaroo, line 369: Maybe this can be `host_signal_cmpxchg!`? Sorry for the bad original suggestion...

src/runtime/signal.jl
```diff
@@ -40,21 +40,20 @@ ROCSignal(signal::HSA.Signal) = ROCSignal(Ref(signal))
...
-function is_done(signal::ROCSignal)
-    value = HSA.signal_load_scacquire(signal.signal[])
-    value < 1
-end
+load_acquire(signal::ROCSignal) = HSA.signal_load_scacquire(signal.signal[])
 
-function get_value(signal::ROCSignal)
-    HSA.signal_load_scacquire(signal.signal[])
-end
+is_done(signal::ROCSignal) = load_acquire(signal) < 1
```
jpsamaroo, line 45: We could instead overload `Base.isdone`

src/runtime/signal.jl
```diff
@@ -40,21 +40,20 @@ ROCSignal(signal::HSA.Signal) = ROCSignal(Ref(signal))
 
 get_handle(signal::ROCSignal) = signal.signal[].handle
 
-function is_done(signal::ROCSignal)
-    value = HSA.signal_load_scacquire(signal.signal[])
-    value < 1
-end
+load_acquire(signal::ROCSignal) = HSA.signal_load_scacquire(signal.signal[])
```
jpsamaroo, line 43: I wonder if we should pull the host signal helpers up from hostcall into the `Runtime` module?

jpsamaroo (commented):
Another quick review of your changes :slightly_smiling_face:

test/runtests.jl
pxl-th: Done

Manifest.toml.gpuarrays-6
pxl-th: Maybe have a separate PR for this?
I'd put moving host signal helpers to `Runtime` module in there as well...

src/device/gcn/hostcall.jl
```diff
@@ -387,8 +346,7 @@ function HostCall(func, rettype, argtypes; return_task=false,
             # the device has read from these buffers. Therefore we wait either for
             # READY_SENTINEL or else an error signal.
             while !Runtime.RT_EXITING[]
-                prev = HSA.signal_load_scacquire(signal.signal[])
-
+                prev = Runtime.get_value(signal)
```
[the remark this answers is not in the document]
jpsamaroo: I think this got missed

src/highlevel.jl
```diff
@@ -102,12 +102,37 @@ function create_executable(device, entry, obj; globals=())
     return ROCExecutable(device, data, entry; globals=globals)
 end
 
+function create_kernel_queue(;
```
jpsamaroo, line 105:
```suggestion
function get_kernel_queue(;
```
We shouldn't actually be creating a new queue here, only getting the one specified by the user.

src/highlevel.jl
```diff
@@ -102,12 +102,37 @@ function create_executable(device, entry, obj; globals=())
...
+    signal::Union{ROCKernelSignal, ROCSignal} = ROCSignal(),
+    device::Union{ROCDevice, Nothing} = nothing,
+    queue::Union{ROCQueue, Nothing} = nothing,
+    kwargs...,
+)
     if signal isa ROCKernelSignal
         return signal
     end
-    return ROCKernelSignal(signal, queue, kernel; kwargs...)
+    kernel_queue = create_kernel_queue(; queue, device)
```
jpsamaroo, line 134:
```suggestion
    kernel_queue = get_kernel_queue(; queue, device)
```

src/highlevel.jl
```diff
@@ -346,7 +372,7 @@ macro roc(ex...)
     args = call.args[2:end]
 
     code = quote end
-    macro_kwargs, compiler_kwargs, call_kwargs, signal_kwargs, kernel_kwargs = split_kwargs(kwargs)
+    device_kw, macro_kwargs, compiler_kwargs, call_kwargs, signal_kwargs, kernel_kwargs = split_kwargs(kwargs)
```
jpsamaroo, line 375:
```suggestion
    device_kwargs, macro_kwargs, compiler_kwargs, call_kwargs, signal_kwargs, kernel_kwargs = split_kwargs(kwargs)
```

src/highlevel.jl
```diff
@@ -390,26 +416,30 @@ macro roc(ex...)
...
                     if $wait
                         foreach($wait!, ($(var_exprs...),))
                     end
                     if $launch
-                        local $kernel_instance = $create_kernel($kernel, $kernel_f, $kernel_args; $(kernel_kwargs...))
-                        local $signal = $create_event($kernel_instance; $(signal_kwargs...))
+                        local $kernel_instance = $create_kernel(
+                            $kernel, $kernel_f, $kernel_args; $(kernel_kwargs...))
+                        local $signal = $create_event(
+                            $kernel_instance; $(device_kw...), $(signal_kwargs...))
```
jpsamaroo, line 442:
```suggestion
                            $kernel_instance; $(device_kwargs...), $(signal_kwargs...))
```

src/highlevel.jl
jpsamaroo: Alternatively, you're welcome to rename these all to `xyz_kw` if you think that's easier to read.

src/runtime/executable.jl
```diff
@@ -70,10 +69,55 @@ end
...
+    if fetch_ex_strings
+        ex_strings = String[]
+        # Check for and collect any exceptions, and clear their slots
+        ex_ring = get_global(exe, :__global_exception_ring)
+        ex_ring_ptr_ptr = Base.unsafe_convert(
+            Ptr{Ptr{AMDGPU.Device.ExceptionEntry}}, ex_ring)
+        ex_ring_ptr = unsafe_load(ex_ring_ptr_ptr)
+
+        while (ex_ring_value = unsafe_load(ex_ring_ptr)).kern != 1
+            if ex_ring_value.kern == signal_handle
```
jpsamaroo, line 106:
```suggestion
            if ex_ring_value.kern == signal_handle && ex_ring_value.ptr != C_NULL
```

I seem to recall that this is sometimes necessary, although that's probably a bug.

src/runtime/execution.jl
```diff
@@ -86,8 +84,7 @@ export roccall
 
 """
     roccall(f::ROCFunction, types, values...;
-            queue::ROCQueue, signal::ROCKernelSignal,
-            groupsize::ROCDim, gridsize::ROCDim)
+            signal::ROCKernelSignal, groupsize::ROCDim, gridsize::ROCDim)
```
jpsamaroo, line 87: This should be:
```julia
roccall(signal::ROCKernelSignal, types, values...; groupsize::ROCDim, gridsize::ROCDim)
```

src/runtime/kernel.jl
```diff
@@ -12,23 +12,43 @@ end
...
+
+    only_handle = signal_handle != UInt64(0)
+    for i in length(m.metadata):-1:1
+        meta = m.metadata[i]
+        if only_handle && (meta.kern == signal_handle)
+            Mem.free(meta.buf)
+            deleteat!(m.metadata, i)
+        else
+            Mem.free(meta.buf)
+        end
```
jpsamaroo, line 33:
```suggestion
        Mem.free(meta.buf)
        if only_handle && (meta.kern == signal_handle)
            deleteat!(m.metadata, i)
        end
```

src/runtime/signal.jl
```diff
@@ -0,0 +1,84 @@
...
+end
+
+struct SignalTimeoutException <: Exception
+    signal::ROCSignal
+end
+
+Adapt.adapt_structure(::Adaptor, sig::ROCSignal) = sig.signal[]
+
+"""
+- `ipc::Bool`: If `true` signal may be used for interprocess communication.
```
jpsamaroo, line 16:
```suggestion
    ROCSignal(init::Integer = 1; ipc::Bool=true) -> ROCSignal

Allocates an HSA signal object which can be used to communicate values between the host and device.

- `ipc::Bool`: If `true`, signal may be used for interprocess communication.
```

src/runtime/signal.jl
```diff
@@ -0,0 +1,84 @@
...
+    IPC signals can be read, written, and waited on from any process.
+"""
+function ROCSignal(init::Integer = 1; interrupt::Bool = true, ipc::Bool = true)
```
jpsamaroo, line 19:
```suggestion
function ROCSignal(init::Integer = 1; ipc::Bool = true)
```

src/runtime/signal.jl
```diff
@@ -0,0 +1,84 @@
...
+            diff_time = (time_ns() - start_time) / 1e9
+            (diff_time > timeout) && throw(SignalTimeoutException(signal))
+        end
+
+        # Allow another scheduled task to run.
+        # This is especially needed in the case
+        # when kernels need to perform HostCalls.
+        yield()
+    end
+    nothing
```
jpsamaroo, line 77:
```suggestion
```

Loops return `nothing` automatically

src/runtime/queue.jl
```diff
@@ -0,0 +1,132 @@
...
+
+"""
+Determine if there are active kernels for the queue.
+If not, we can re-use it.
+"""
+function has_active_kernels(q::ROCQueue)
+    lock(RT_LOCK) do
+        active_kernels = get(_active_kernels, q, nothing)
+        isnothing(active_kernels) && return false
+        !isempty(active_kernels)
```
jpsamaroo, line 104:
```suggestion
        return !isempty(active_kernels)
```

src/runtime/queue.jl
```diff
@@ -0,0 +1,132 @@
...
+    end
+end
+
+function kill_queue!(queue::ROCQueue)
+    (@atomic queue.active) || return nothing
+    @atomic queue.active = false
```
jpsamaroo, line 110: This should use `@atomicreplace` or `@atomicswap` to avoid TOCTTOU races:

```julia
if !(@atomicswap queue.active = false)
    return
end
```

src/runtime/queue.jl
```diff
@@ -0,0 +1,132 @@
...
+
+        # Send exception to all waiter signals.
+        if queue.status != HSA.STATUS_SUCCESS
+            err = QueueError(queue.queue, HSAError(queue.status))
+            for kersig in _active_kernels[queue]
+                kersig.exception = err
+                notify(kersig)
+            end
+        end
+        # Delete queue from global holders.
```
jpsamaroo, line 126:
```suggestion
        # Clear active kernel set
```

src/runtime/queue.jl
```diff
@@ -0,0 +1,132 @@
...
+
+function kill_queue!(queue::ROCQueue)
+    (@atomic queue.active) || return nothing
+    @atomic queue.active = false
+
+    lock(RT_LOCK) do
+        if get(DEFAULT_QUEUES, queue.device, nothing) == queue
+            delete!(DEFAULT_QUEUES, queue.device)
+        end
+        delete!(QUEUES, queue.queue)
```
jpsamaroo, line 116: We should move this last line (deleting from `QUEUES`) directly into the finalizer to ensure that it only gets removed once there are no more references to this `ROCQueue` object (otherwise async code might get a `KeyError`).

src/runtime/kernel-signal.jl
```diff
@@ -0,0 +1,90 @@
...
+        isnothing(kersig.exception) && (kersig.exception = err;)
+        false
+    finally
+        queue_error = kersig.queue.status != HSA.STATUS_SUCCESS
+        # If it is QueueError, `kill_queue!` will fill `exception` field.
+        queue_error && kill_queue!(kersig.queue)
+    end
+
+    cleanup!(kersig; finished, check_exceptions)
+    finished
```
jpsamaroo, line 53:
```suggestion
    return finished
```

src/runtime/kernel-signal.jl
```diff
@@ -0,0 +1,90 @@
...
+end
+
+function cleanup!(
+    kersig::ROCKernelSignal; finished::Bool, check_exceptions::Bool = true,
+)
+    (@atomic kersig.active) || return nothing
+    @atomic kersig.active = false
```
jpsamaroo, line 60:
```suggestion
    if !(@atomicswap kersig.active = false)
        return
    end
```

Avoid TOCTTOU race

src/runtime/kernel-signal.jl
```diff
@@ -0,0 +1,90 @@
...
+
+    unpreserve!(kersig)
+    finalize(kersig.kernel)
```
jpsamaroo, line 63:
```suggestion
```
We don't know that all references are gone at this point (and in fact, they probably aren't, since this function inherently preserves the object until the end of its execution).

src/runtime/kernel-signal.jl
```diff
@@ -0,0 +1,90 @@
...
+
+    exe::ROCExecutable = kersig.kernel.exe
+    mod::ROCModule = EXE_TO_MODULE_MAP[exe].value
+    signal_handle::UInt64 = (kersig.signal.signal[]::HSA.Signal).handle
```
jpsamaroo, line 67:
```suggestion
    signal_handle::UInt64 = get_handle(kersig.signal)
```

jpsamaroo (approved):
Awesome stuff!

src/runtime/queue.jl
pxl-th: I think we should use `@atomicreplace` to ensure the queue can only be killed once.
We check that `queue` is active, and only when it is `true` we set it to `false` and continue executing the function:
```julia
  _, succ = @atomicreplace queue.active true => false
  succ || return
  ...
```

src/runtime/queue.jl
pxl-th: Having it in finalizer brings back this bug: [link](https://github.com/JuliaGPU/AMDGPU.jl/pull/265#issuecomment-1255550402).

It looks like HSA might reuse address for the newly created queue, if the previous one was deleted.
If the queue is killed, shouldn't we ensure it is not accessible from `QUEUES` as well, so that we won't trigger `@assert !haskey(QUEUES, queue.queue)` statement during queue creation?
It is killed so not valid, at least that specific instance of `ROCQueue`.

`kill_queue!` is called from finalizer as well, so once there are no references to it, it will be removed from `QUEUES`.

jpsamaroo: Ok, this makes sense to me. We don't currently use `QUEUES` for anything other than that assertion, so let's just leave it (or feel free to remove it entirely).

Issue #305, opened 2022-10-05
Segfault in array initialisation


luraess:
Using AMDGPU 0.4.3 segfaults upon array initialisation `AMDGPU.ones(Float64,2,2)` or `ROCArray(ones(2,2))`. Also, it is unclear to me why it now uses `hipStreamSynchronize`. (@jpsamaroo or @vchuravy do you have any insights on what's going on here?)

This occurs using Julia 1.8.2 and ROCm 4.3 on a system with Vega20 (gfx906) cards using artifacts.

It looks like that `JULIA_AMDGPU_DISABLE_ARTIFACTS` has no longer any effect. 

_Testing with AMDGPU 0.4.2 all works fine, and also env var to disable artifacts works._

```julia
julia> AMDGPU.versioninfo()
HSA Runtime (ready)
- Path: /users/lraess/.julia/artifacts/20030eaff1f8f47d3646fc99d415a823516778d7/lib/libhsa-runtime64.so
- Version: 1.1.0
ld.lld (ready)
- Path: /users/lraess/.julia/artifacts/c86785d1da1b021c7790274eb700100581e341a5/tools/lld
ROCm-Device-Libs (ready)
- Path: /users/lraess/.julia/artifacts/2a4556d4ad40fc77472f80ad8a090d3ffea854bb/amdgcn/bitcode
HIP Runtime (ready)
- Path: /users/lraess/.julia/artifacts/5c755583cc986bc9e32bd640cdb0045f19a094bd/hip/lib/libamdhip64.so
rocBLAS (ready)
- Path: /users/lraess/.julia/artifacts/1fd1782850f6482c5bdbeec69cb3d1e43d949a3c/lib/librocblas.so
rocSOLVER (MISSING)
rocALUTION (MISSING)
rocSPARSE (ready)
- Path: /users/lraess/.julia/artifacts/edcc84addc716a5f1f4ee7af62643e616ce23886/rocsparse/lib/librocsparse.so
rocRAND (ready)
- Path: /users/lraess/.julia/artifacts/6f32e80b256ed82bae6982288b859e244a36479d/lib/librocrand.so
rocFFT (MISSING)
MIOpen (MISSING)
HSA Agents (6):
- CPU-XX [AMD EPYC 7742 64-Core Processor]
- CPU-XX [AMD EPYC 7742 64-Core Processor]
- GPU-c616498172e626c4 [gfx906]
- GPU-ebae40e172e620f6 [gfx906]
[...]

PR #307, opened 2022-10-10
Update bindeps search path

luraess:
Update the binders code to also search for `ROCM_PATH` as suggested by @jpsamaroo . This is needed to build AMDGPU using system libs on LUMI with Julia 1.8.2.


luraess:
This mechanism helps in finding the libraries but does not prevent from getting a segfault as reported in #308 (potentially related to #305 as well).

jpsamaroo:
Thanks!

PR #311, opened 2022-10-14
Add dynamic local memory support

### LINKED ISSUE #304: Add AMDGPU.jl equivalent to CUDA.CuDynamicSharedArray (and CUDA.CuStaticSharedArray)
Would it be possible to add an equivalent to [`CUDA.CuDynamicSharedArray`](https://cuda.juliagpu.org/stable/api/kernel/#CUDA.CuDynamicSharedArray) (a `CUDA.CuStaticSharedArray` equivalent we don't need, but would certainly be expected to be available by some users)?

jpsamaroo:
Allows dynamically allocating local memory at runtime, by doing something like:

```julia
function f(RA::ROCDeviceArray)
    ptr = AMDGPU.Device.alloc_local(:my_shmem, Float32, 0) # 0 signals dynamic allocation
    RB = ROCDeviceArray(length(RA), ptr)
end
```

Closes #304

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 *.jl.mem
 
 docs/build
+LocalPreferences.toml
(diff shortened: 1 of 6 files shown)


omlins:
Would it be possible to wrap that in a single function or macro call, analog to [CUDA.CuDynamicSharedArray](https://cuda.juliagpu.org/stable/api/kernel/#CUDA.CuDynamicSharedArray) or the old `CUDA.@cuDynamicSharedMem` macro?

jpsamaroo:
Yep, that should be easy enough (and I really should have just added it here).

omlins:
> Yep, that should be easy enough (and I really should have just added it here).

@jpsamaroo Could you implement this macro or function? How is it called?

Issue #277, opened 2022-09-05
Consolidate thread lock/data race handling


torrance:
I've noticed a number of competing techniques used to manage data races throughout the codebase.

There is the RT_LOCK (I assume RT = runtime) used to manage global state access and modifications. However, it's not immediately clear _which_ objects require this RT_LOCK, and its use is enforced only by code review.

In a few other places I've also noticed use of atomics to handle these cases, for example in the Queue objects during deletion.

I would like to suggest standardising on how we handle this to ensure correctness and simplicity of design. It makes code brittle if their are competing paradigms for handling data race conditions.

I'm not an expert in how to handle this kind of thing, and I'm sure there are various options but I'd like to propose to couple any shared state object with its lock, e.g. something like:

```
struct LockedObject{T}
    lock Threads.ReentrantLock
    payload T
end
```
and accesses to this object can happen via a pattern such as:
```
lock(QUEUES::LockedObject) do QUEUES
    # push or pop or whatever
end
```
where `lock()` is simply something like:
```
function lock(f, x)
    lock(x.lock)
    try
        return f(x.payload)
    finally
        unlock(x.lock)
    end
end
```

The benefits here mean its obvious which objects require locking before use, and any collection of objects which require synchronisation can share a lock between them.

My 2 cents. But I think it's a discussion that needs to happen.

jpsamaroo:
Yeah, this is a great point. My intention for `RT_LOCK` is for it to be a simple way to make access to various global objects safe (especially anything that's an `AbstractDict`). But we don't really want a big global lock for a lot of things, so having a `LockedObject` concept sounds like a great approach.

The queue atomics (on `.queue` and `.active`) were put in in an attempt to fix the various queue issues that we had been seeing in CI (most of which I think had other solutions). While I do think one of them should remain atomic (so that we can, without lock contention, check queue state just before kernel launch or queue queries), the other can and should probably become not-atomic or be removed. I would personally want to see `.active` disappear and have us check if `.queue == C_NULL` to check if a queue is live (and we'd then atomically store `C_NULL` to it on queue error).

pxl-th:
Closing this as we now use [`LockedObject`](https://github.com/JuliaGPU/AMDGPU.jl/blob/e57a7fb8021550b6c56297767b76f74ee5b0ec44/src/AMDGPU.jl#L39) throughout the codebase for this.

PR #320, opened 2022-11-02
Add MIOpen

pxl-th:
This PR adds initial support for MIOpen.
Mainly it will target only convolutions for now.

- Add auto-generated MIOpen wrapper.
- Add tensor/convolution descriptors.
- Implement high-level convolution functions.
- Implement benchmark caching.
- Add MIOpen tests.

Example of convolution API:
```julia
x = AMDGPU.ones(Float32, 3, 3, 1, 1)
w = AMDGPU.ones(Float32, 3, 3, 1, 1)
y = MIOpen.convolution(x, w; padding=(0, 0), stride=(1, 1), dilation=(1, 1), groups=1)
```

**TODO**
- [x] Complete convolution wrappers.
- [x] Add more tests.
- [x] Update docs.
- [x] Investigate the source of errors when following env variables are not set: `MIOPEN_DEBUG_CONV_DIRECT_NAIVE_CONV_FWD=0`, `MIOPEN_DEBUG_CONV_DIRECT_NAIVE_CONV_BWD=0`, `MIOPEN_DEBUG_CONV_DIRECT_NAIVE_CONV_WRW=0`.
Potential fix is to rebuild HIP (hipamd) with PCH `on`.
UPD: Should be fixed with this PR: https://github.com/JuliaPackaging/Yggdrasil/pull/5797
- [x] Update CI.

---

Code to generate MIOpen wrapper using Clang.jl

```julia
using Clang.Generators
using MIOpen_jll

include_dir = normpath(MIOpen_jll.artifact_dir, "include")
miopen_dir = joinpath(include_dir, "miopen")
options = load_options("generator.toml")

args = get_default_args()
push!(args, "-I$include_dir")

headers = [
    joinpath(miopen_dir, header)
    for header in readdir(miopen_dir)
    if endswith(header, ".h")
]

ctx = create_context(headers, args, options)
build!(ctx)
```

```toml
[general]
module_name = "libMIOpen"
library_name = "libMIOpen_path"
output_file_path = "./libMIOpen.jl"
jll_pkg_name = "MIOpen_jll"
export_symbol_prefixes = []
```


pxl-th:
At this moment it supports 2D & 3D convolutions.
Supported types are `Float16` & `Float32`.

For other types I get errors that look like either misconfigured `MIOpen_jll` or bug in MIOpen itself, not sure :/
I suspect both, maybe ROCm 5.3 - 5.4 will have more things working...

pxl-th:
@jpsamaroo not sure what to do about CI.
It fails because MIOpen is not present in `Manifest.toml` and because it was built only for Julia 1.9.

To fix buildkite CI we can delete `Manifest.toml`.
And for GitHub CI we can upgrade it to 1.9...
Although that leaves out Julia 1.7...

Project.toml
```diff
@@ -17,6 +17,7 @@ LLVM = "929cbde3-209d-540e-8aea-75f648917ca0"
 LLVM_jll = "86de99a1-58d6-5da7-8064-bd56ce2e322c"
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+MIOpen_jll = "2409bb75-d5ef-542a-ac68-1cfd4c37dc24"
```
jpsamaroo, line 20: Needs a compat entry

deps/bindeps.jl
```diff
@@ -181,7 +181,7 @@ function bindeps_setup()
         :device_libs_build_reason => "unknown",
     )
     for (name, _) in rocm_ext_libs
-        lib = Symbol(:lib, lowercase(string(name)))
+        lib = Symbol(:lib, name == :MIOpen ? string(name) : lowercase(string(name)))
```
jpsamaroo, line 184:
```suggestion
        lib = Symbol(:lib, name)
```

Since the others are already lowercase. I think the `lowercase` call was just a relic of older code.

deps/bindeps.jl
```diff
@@ -359,7 +359,7 @@ function bindeps_setup()
         ### Find HIP-based libraries
         @sync for (name, pkg) in rocm_ext_libs
             @async begin
-            lib = Symbol(:lib, lowercase(string(name)))
+            lib = Symbol(:lib, name == :MIOpen ? string(name) : lowercase(string(name)))
```
jpsamaroo, line 362:
```suggestion
            lib = Symbol(:lib, name)
```
As above

test/runtests.jl
```diff
@@ -116,6 +116,13 @@ push!(tests, "NMF" => ()->begin
         @test_skip "NMF"
     end
 end)
+push!(tests, "MIOpen" => ()->begin
+    if AMDGPU.functional(:MIOpen)
+        include("dnn/miopen.jl")
```
jpsamaroo, line 121: I like the move from the `rocarray` directory to a dedicated `dnn` directory! For future PRs to the other libraries (rocBLAS et. al), feel free to also move them to their own folders, for consistency.

test/dnn/miopen.jl
```diff
@@ -0,0 +1,90 @@
+using AMDGPU.MIOpen
+
+const MIOPEN_TYPES = (
+    (Float16, MIOpen.miopenHalf),
+    (Float32, MIOpen.miopenFloat),
+    (Float64, MIOpen.miopenDouble),
+    (Int8, MIOpen.miopenInt8),
+    (Int32, MIOpen.miopenInt32),
+)
+const MIOPEN_CONV_TYPES = (Float16, Float32) # FIXME: support more types
```
jpsamaroo, line 10: Maybe we can move these type definitions into AMDGPU.jl, instead of just defining them in the tests?

test/dnn/miopen.jl
```diff
@@ -0,0 +1,90 @@
...
+        @test eltype(y) == T
+        @test size(y) == (target_size..., 1, 1)
+        @test Array(y)[1] == prod(spatial_size)
+
+        Δ = AMDGPU.ones(T, size(y))
+        ∇w = MIOpen.∇convolution_weight(Δ, x, w; padding, stride, dilation, groups)
+        @test size(∇w) == size(w)
+
+        ∇x = MIOpen.∇convolution_data(Δ, x, w; padding, stride, dilation, groups)
+        @test size(∇x) == size(x)
```
jpsamaroo, line 52: Is there any easy way to check that these results are correct? Like how with the rocBLAS tests, we compare against CPU BLAS.

test/dnn/miopen.jl
```diff
@@ -0,0 +1,90 @@
...
+
+    # Depthwise convolution.
+    wd1 = AMDGPU.rand(Float32, 2, 2, 1, 3)
+    y = MIOpen.convolution(
+        x, wd1; padding=(0, 0), stride=(1, 1), dilation=(1, 1), groups=3)
+    @test size(y) == (9, 9, 3, 10)
+
+    x = AMDGPU.ones(Float32, 10, 10, 4, 10)
+    wd2 = AMDGPU.ones(Float32, 2, 2, 2, 4)
+    y = AMDGPU.MIOpen.convolution(
```
jpsamaroo, line 87:
```suggestion
    y = MIOpen.convolution(
```

src/dnn/MIOpen.jl
```diff
@@ -0,0 +1,77 @@
+module MIOpen
+
+import AMDGPU
+import AMDGPU.Runtime.Mem
+
+using AMDGPU: ROCArray, ROCDevice
```
jpsamaroo, line 6:
```suggestion
import .AMDGPU: ROCArray, ROCDevice
```

I think this is equivalent? I like to avoid `using` unless we need it to get exported things.

src/dnn/MIOpen.jl
```diff
@@ -0,0 +1,77 @@
...
+    if     status == miopenStatusSuccess return "Success"
+    elseif status == miopenStatusNotInitialized return "Not initialized"
+    elseif status == miopenStatusInvalidValue return "Invalid value"
+    elseif status == miopenStatusBadParm return "Bad parameter"
+    elseif status == miopenStatusAllocFailed return "Allocation failed"
+    elseif status == miopenStatusInternalError return "Internal error"
+    elseif status == miopenStatusNotImplemented return "Not implemented"
+    elseif status == miopenStatusUnknownError return "Unknown error"
+    elseif status == miopenStatusUnsupportedOp return "Unsupported operation"
+    elseif status == miopenStatusGpuOperationsSkipped return "GPU operations skipped" end
```
jpsamaroo, line 27: It's usually my preference to put these mappings into a `Dict`, and then `get_status_string` can do a lookup from it. I know that this `if ... elseif` logic probably makes for faster code, but it's more verbose and doesn't allowing checking (without throwing and catching an error) if a certain status code has an associated status string.

src/dnn/MIOpen.jl
```diff
@@ -0,0 +1,77 @@
...
+    error("Invalid status code: $status")
+end
+
+function check(status)
+    if status != miopenStatusSuccess
+        status_string = get_status_string(status)
+        error("MIOpen return status [code=$status]: $status_string")
```
jpsamaroo, line 34: Maybe we can make an `MIOpenException` object (which contains the numeric status code) and throw that instead, so that it's easier for libraries to catch MIOpen errors? Then this string can be part of the `Base.showerror` implementation.

src/dnn/MIOpen.jl
```diff
@@ -0,0 +1,77 @@
...
+    end
+end
+
+function get_miopen_data_type(t)
+    if     t == Float16 return miopenHalf
+    elseif t == Float32 return miopenFloat
+    elseif t == Float64 return miopenDouble
+    elseif t == Int8    return miopenInt8
+    elseif t == Int32   return miopenInt32 end
```
jpsamaroo, line 43: Same as for `get_status_string`, put these in a `Dict`.

src/dnn/MIOpen.jl
```diff
@@ -0,0 +1,77 @@
...
+    error("Unsupported data type: $t")
+end
+
+function get_version()
```
jpsamaroo, line 47:
```suggestion
function version()
```

src/dnn/MIOpen.jl
```diff
@@ -0,0 +1,77 @@
...
+end
+
+function check(status)
+    if status != miopenStatusSuccess
+        status_string = get_status_string(status)
+        error("MIOpen return status [code=$status]: $status_string")
+    end
+end
+
+function get_miopen_data_type(t)
```
jpsamaroo, line 38:
```suggestion
function miopen_data_type(t)
```

src/dnn/MIOpen.jl
```diff
@@ -0,0 +1,77 @@
...
+using GPUArrays
+using MIOpen_jll
+
+const libMIOpen_path = MIOpen_jll.libMIOpen_path
+
+include("low_level.jl")
+
+const HANDLE = Ref{miopenHandle_t}(C_NULL)
+
+function get_status_string(status)
```
jpsamaroo, line 17:
```suggestion
function status_string(status)
```

src/dnn/MIOpen.jl
```diff
@@ -0,0 +1,77 @@
...
+    miopenCreate(handle) |> check
+    handle[]
+end
+
+function destroy_handle!(handle::miopenHandle_t)
+    miopenDestroy(handle) |> check
+    nothing
+end
+
+function get_handle()
```
jpsamaroo, line 64:
```suggestion
function handle()
```

src/dnn/MIOpen.jl
```diff
@@ -0,0 +1,77 @@
...
+    global HANDLE
```
jpsamaroo, line 65:
```suggestion
```

I don't think you need this, since it's already a global and doesn't conflict with other names.

src/dnn/MIOpen.jl
```diff
@@ -0,0 +1,77 @@
...
+    if HANDLE[] == C_NULL
+        handle = create_handle()
+        atexit(() -> destroy_handle!(handle))
+        HANDLE[] = handle
+    end
+    HANDLE[]
```
jpsamaroo, line 71: This has a TOCTTOU race; we should consider using something like what @torrance suggested in https://github.com/JuliaGPU/AMDGPU.jl/issues/277

src/dnn/convolution.jl
```diff
@@ -0,0 +1,282 @@
...
+        expand(stride, 3), expand(dilation, 3), groups)
+end
+
+const CONV_FWD_BENCHMARK_CACHE = Dict{ConvolutionArgs, miopenConvAlgoPerf_t}()
+const CONV_BWD_WEIGHT_BENCHMARK_CACHE = Dict{ConvolutionArgs, miopenConvAlgoPerf_t}()
+const CONV_BWD_DATA_BENCHMARK_CACHE = Dict{ConvolutionArgs, miopenConvAlgoPerf_t}()
+
+get_conv_cache_type(::Type{miopenConvFwdAlgorithm_t}) = CONV_FWD_BENCHMARK_CACHE
+get_conv_cache_type(::Type{miopenConvBwdDataAlgorithm_t}) = CONV_BWD_DATA_BENCHMARK_CACHE
+get_conv_cache_type(::Type{miopenConvBwdWeightsAlgorithm_t}) = CONV_BWD_WEIGHT_BENCHMARK_CACHE
```
jpsamaroo, line 38: We need to consider locking here.

src/dnn/convolution.jl
```diff
@@ -0,0 +1,282 @@
...
+    _check_conv_args_dims(K, N)
+    xdesc, wdesc, cdesc, pi32, si32, di32 = _get_conv_descriptors(
+        x, w; padding, stride, dilation, groups)
+    y = similar(x, T, get_output_size(cdesc, xdesc, wdesc))
+    convolution!(
+        y, TensorDescriptor(y), x, xdesc, w, wdesc, cdesc;
+        padding=pi32, stride=si32, dilation=di32, groups)
+end
+
+function ∇convolution_weight!(
```
jpsamaroo, line 186: I'm ok with using Unicode in the names for these functions, but I would also like to add another function like `delta_convolution_weight!` or similar, which forwards to `∇convolution_weight!`, for users who can't easily use Unicode with their keyboards.

src/dnn/descriptors.jl
```diff
@@ -0,0 +1,91 @@
+const MIOPEN_DIM_MAX = 5
+
+function check_ndims(x)
+    if ndims(x) > MIOPEN_DIM_MAX
+        error(
```
jpsamaroo, line 5: Could this be its own exception type?

src/dnn/descriptors.jl
```diff
@@ -0,0 +1,91 @@
...
+    TensorDescriptor(dtype, ndims(x), sizes, strides_)
+end
+
+function Base.ndims(desc::TensorDescriptor)
+    nd = Ref{Int32}(0)
+    miopenGetTensorDescriptorSize(desc.handle, nd) |> check
+    Int64(nd[])
+end
+
+function get(desc::TensorDescriptor)
```
jpsamaroo, line 48: Can we rename this to something like `unpack` or `params` or similar? `get` is somewhat ambiguous in what it actually gets.

jpsamaroo (commented):
Awesome work! My major requests are:
- Implementing locking around global data access
- Removing `get_` prefixes

test/dnn/miopen.jl
pxl-th: Easiest would be to use NNlib. Should we do that?

pxl-th: Added!

ToucheSir:
In case you haven't seen it already, the CUDA.CUDNN [readme](https://github.com/JuliaGPU/CUDA.jl/tree/master/lib/cudnn) may be useful for discussions on function naming. That and the test code in the subpackage, in case you haven't adapted it for AMDGPU.jl already.

pxl-th:
> In case you haven't seen it already, the CUDA.CUDNN [readme](https://github.com/JuliaGPU/CUDA.jl/tree/master/lib/cudnn) may be useful for discussions on function naming. That and the test code in the subpackage.

No I haven't seen that.
For the high-level interface I was actually aiming at being more similar to NNlib's API.
So that it is easy to use it even without NNlib.

pxl-th:
Regarding the error in CI:
```
From worker 4:	MIOpen(HIP): Error [Do] 'amd_comgr_do_action(kind, handle, in.GetHandle(), out.GetHandle())' AMD_COMGR_ACTION_COMPILE_SOURCE_TO_BC: ERROR (1)
--
  | From worker 4:	MIOpen(HIP): Error [BuildHip] comgr status = ERROR (1)
  | From worker 4:	MIOpen(HIP): Warning [BuildHip] <built-in>:1:10: fatal error: '__clang_hip_runtime_wrapper.h' file not found
  | From worker 4:	#include "__clang_hip_runtime_wrapper.h"
  | From worker 4:	         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  | From worker 4:	1 error generated when compiling for gfx90a.
```

I dont have it on gfx1031.
My guess is that it could be a misconfigured MIOpen, possible solution to try: https://github.com/cupy/cupy/issues/5592#issuecomment-901136342
Or a bug in MIOpen itself...
I plan to update ROCm stack to either 5.3.x or 5.4 in the nearest future as it contains several improtant improvements, which might help with things.

pxl-th:
> Regarding the error in CI:
> 
> ```
> From worker 4:	MIOpen(HIP): Error [Do] 'amd_comgr_do_action(kind, handle, in.GetHandle(), out.GetHandle())' AMD_COMGR_ACTION_COMPILE_SOURCE_TO_BC: ERROR (1)
> --
>   | From worker 4:	MIOpen(HIP): Error [BuildHip] comgr status = ERROR (1)
>   | From worker 4:	MIOpen(HIP): Warning [BuildHip] <built-in>:1:10: fatal error: '__clang_hip_runtime_wrapper.h' file not found
>   | From worker 4:	#include "__clang_hip_runtime_wrapper.h"
>   | From worker 4:	         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
>   | From worker 4:	1 error generated when compiling for gfx90a.
> ```
> 
> I dont have it on gfx1031. My guess is that it could be a misconfigured MIOpen, possible solution to try: [cupy/cupy#5592 (comment)](https://github.com/cupy/cupy/issues/5592#issuecomment-901136342) Or a bug in MIOpen itself... I plan to update ROCm stack to either 5.3.x or 5.4 in the nearest future as it contains several improtant improvements, which might help with things.

CI job that does not use artifacts fails with the same error. So it might not be artifact's fault.

Project.toml
```diff
@@ -44,6 +45,7 @@ HIP_jll = "4, 5"
 LLD_jll = "12, 13, 14"
 LLVM = "4.6"
 LLVM_jll = "12, 13, 14"
+MIOpen_jll = "5"
```
jpsamaroo, line 48:
```suggestion
MIOpen_jll = "4, 5"
```
And an associated JLL needs to exist, otherwise Julia 1.7 won't be able to load AMDGPU due to resolver errors.

jpsamaroo (approved):
Awesome!

Project.toml
pxl-th: I'll look into adding MIOpen for 4.2.0 if it is not causing too many issues.

PR #321, opened 2022-11-05
Add support for 2D * 3D batched GEMM

pxl-th:
This PR adds support for 2D * 3D batched GEMM.
One use-case is applying the same `4x4` intrinsics matrix `K` to different extrinsics `4x4xN` matrices `[R|t]`.


pxl-th:
I think it is safe to merge it.

PR #322, opened 2022-11-07
Support NNlib batched gemm format

pxl-th:
Turns out, NNlib internally reshapes 2D arrays to 3D and passes only 3D arrays to actual batched gemm implementations.
So this PR adjusts to that and drops 2D * 3D batched gemm support for simplicity.


pxl-th:
Should be safe to merge.

Issue #319, opened 2022-11-01
rocBLAS GEMM ignores `@view`


pxl-th:
Probably just need to pass correct buffer pointers to gemm...

```julia
julia> using AMDGPU

julia> x = AMDGPU.rand(Float32, 4, 4, 3);

julia> @view(x[:, :, 3]) * @view(x[:, :, 3]) # Incorrect result.
4×4 ROCMatrix{Float32}:
 0.860865  0.52898   0.622481  1.4993
 1.23489   0.932102  1.54347   2.36086
 0.516221  0.446005  0.913586  1.21474
 0.418985  0.309178  0.497969  1.0506

julia> x[:, :, 3] * x[:, :, 3] # This is correct result.
4×4 ROCMatrix{Float32}:
 0.417364  1.27887   1.05345   1.48215
 0.25803   0.933164  0.802089  1.11318
 0.368032  1.04568   0.900236  1.05582
 0.284477  0.53608   0.647899  0.542196

julia> x[:, :, 1] * x[:, :, 1] # This is what is actually computed.
4×4 ROCMatrix{Float32}:
 0.860865  0.52898   0.622481  1.4993
 1.23489   0.932102  1.54347   2.36086
 0.516221  0.446005  0.913586  1.21474
 0.418985  0.309178  0.497969  1.0506
```

torrance:
I've spent the last few days debugging a problem that comes down to the same issue, except with FFT.

For example:

```julia
@testset "FFT and view" begin
    As = rand(Complex{Float64}, 96, 96, 2)
    As_d = ROCArray(As)

    A = view(As, :, :, 1)
    A_d = view(As_d, :, :, 1)

    @test A == Array(A_d)  # passes

    Pigi.fft!(A)
    Pigi.fft!(A_d)

    @test As == Array(As_d)  # fails
end
```

So it seems ROCArrays which are actually views of an underlying array are not correctly passed to roc libraries.

PR #323, opened 2022-11-08
Add pointer() method for ROCArray and some library tests

torrance:
This PR mean to resolve [issue](https://github.com/JuliaGPU/AMDGPU.jl/issues/319) by providing a `Base.pointer(::ROCArray)` method that correctly handles non-zero offsets.

(Honestly, I don't understand how the calls to `pointer(::ROCArray)` were working at all before??!)

Currently, the FFT test now passes, however, the the BLAS test still fails. Still attempting to work out why `pointer()` is not being called when performing matrix multiplication.

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -85,8 +85,6 @@ function unsafe_free!(xs::ROCArray)
     return
 end
 
-Base.pointer(x::ROCArray) = x.buf.ptr + x.offset
-
 wait!(x::ROCArray) = wait!(x.syncstate)
 mark!(x::ROCArray, s) = mark!(x.syncstate, s)
 wait!(xs::Vector{<:ROCArray}) = foreach(wait!, xs)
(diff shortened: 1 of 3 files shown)


pxl-th:
> Still attempting to work out why pointer() is not being called when performing matrix multiplication.

Probably because we pass raw buffer pointer (`x.buf`) instead of calling `pointer(x)`

torrance:
Ok, so I understand now how this was working in the first place: the calls to `pointer()` were ultimately calling `unsafe_convert(Ptr{T}, ::ROCArray{T})`. Fixing this function instead means explicit calls to `pointer()` as well as implicit conversions to pointer as part of `ccall()` now work.

This also resolves the BLAS bug.

I think this should be ready to merge. @jpsamaroo can you review this small change?

pxl-th:
We probably should also check (and error?) if the view results in non-contiguous array, like following:

```julia
julia> x = AMDGPU.rand(Float32, 16, 16, 2);

julia> @view(x[4:8, 4:8, 1]) * @view(x[4:8, 4:8, 1])
ERROR: ReadOnlyMemoryError()
```

and allow views that only offest from the beginning.

jpsamaroo:
Thanks @torrance ! @pxl-th let's add that in a separate PR.

PR #325, opened 2022-11-15
Mem: Allow using hipMalloc/hipFree for allocations

jpsamaroo:
By setting JULIA_AMDGPU_USE_HIP_MALLOC_OVERRIDE=1, or calling `AMDGPU.Runtime.Mem.enable_hip_malloc_override!()`, AMDGPU will instead use HIP to allocate and free memory to back `ROCArray` allocations. This will be necessary for the upcoming rocSPARSE support, as rocSPARSE requires HIP allocations for some operations.

This is an experimental option and is subject to future removal if and when HIP allocations are no longer necessary for ROCm library calls.


jpsamaroo:
I'm going to leave this undocumented and untested for right now, since I think we want to have more full-fledged (and semantically correct) HIP integration before advertising it. Specifically, selecting the right device to allocate/free on, and generally having a notion of per-task HIP contexts and streams, like CUDA.jl.

PR #326, opened 2022-11-15
Add occupancy API and groupsize tuning

### LINKED ISSUE #271: Implement occupancy API
CUDA has a great feature for sizing threads and blocks, namely launch_configuration(). I rarely manually size my kernel, instead something like:

```
kernel = @cuda launch=false myfunc(args...)
config = launch_configuration(kernel.fun)
threads = min(N, config.threads)
blocks = cld(N, threads)
kernel(args...; threads, blocks)
```

It’s almost always very close to optimal, and allows my code to move from device to device without worrying too much about launch parameters.

ROCM provides something similar in https://github.com/ROCm-Developer-Tools/hipamd/blob/3ec1ccdbbbee7090ba854eddd1dee281973a4498/src/hip_platform.cpp#L301

All array level kernels ultimately rely on GPUArray's `launch_configuration()`, which currently defaults to `threads=256, blocks=32`, and which is suboptimal. Enabling this functionality would improve all array-level programming.

This issue tracks the work required to port this function into AMDGPU.jl

jpsamaroo:
Implements an occupancy API (borrowed from HIP), and uses it to enable `AMDGPU.launch_configuration` like CUDA, as well as allowing `groupsize=:auto` to enable automatic groupsize selection.

Closes #271

Todo:
- [x] Move occupancy code to compiler
- [x] Change `localmem` semantics to mean dynamic local memory
- [x] Propagate `localmem` to `roccall` for `groupsize=:auto`
- [x] Test that occupancy is invalidated upon recompilation
- [x] Make `calculate_occupancy` nicer for manual usage
- [x] Add tests and docs


src/runtime/execution.jl
```diff
@@ -65,11 +65,12 @@ end
 
 @doc (@doc AbstractKernel) HostKernel
 
-@inline function roccall(kernel::HostKernel, tt, args...; signal::ROCKernelSignal, config=nothing, kwargs...)
-    if config !== nothing
-        roccall(signal, tt, args...; config(kernel)..., kwargs...)
+@inline function roccall(kernel::HostKernel, tt, args...; signal::ROCKernelSignal, groupsize=1, kwargs...)
+    if groupsize == :auto
+        config = AMDGPU.launch_configuration(kernel; dynamic_localmem=localmem)
```
pxl-th, line 70: `localmem` is undefined. We need to pass through `kernel_kwargs` from `@roc`.

src/highlevel.jl
```diff
@@ -452,3 +454,12 @@ macro roc(ex...)
         end
     return esc(code)
 end
+
+# launch config
+
+function launch_configuration(fun::Runtime.ROCFunction, device::ROCDevice; input_block_size=1, dynamic_localmem=0)
```
pxl-th, line 460: Maybe we should get the `device` from `fun` as `fun.mod.exe.device`?
To avoid ambiguity...

```suggestion
function launch_configuration(fun::Runtime.ROCFunction; input_block_size=1, dynamic_localmem=0)
```

jpsamaroo:
Thanks for the help @pxl-th !

PR #328, opened 2022-11-21
Cast to Ptr before checking NULL pointer

torrance:
The LLVM Ptr can't be directly compared to the C_NULL pointer - it needs to first be cast.

(Be warned, I have no idea what this code is doing, only that I encountered this bug during testing.)


jpsamaroo:
Nice catch, thanks! TIL :smile:

PR #329, opened 2022-11-21
CI: Re-enable Julia 1.7 CI

jpsamaroo:


jpsamaroo:
Apparently Helios also has issues with tests hanging, so I'll take them back offline until I can figure that out.

pxl-th:
Since #423, we now support only Julia 1.9+.

PR #333, opened 2022-11-28
Resize! support

matinraayai:
Adds support for resize! for ROCVector.

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -347,7 +347,7 @@ GPUArrays.backend(::Type{<:ROCArray}) = ROCArrayBackend()
 
 function Base.convert(::Type{ROCDeviceArray{T,N,AS.Global}}, a::ROCArray{T,N}) where {T,N}
     ptr = Base.unsafe_convert(Ptr{T}, a.buf)
-    ROCDeviceArray{T,N,AS.Global}(a.dims, AMDGPU.LLVMPtr{T,AS.Global}(ptr+a.offset))
+    ROCDeviceArray{T,N,AS.Global}(a.dims, AMDGPU.LLVMPtr{T,AS.Global}(ptr + a.offset))
 end
 Adapt.adapt_storage(::Runtime.Adaptor, x::ROCArray{T,N}) where {T,N} =
     convert(ROCDeviceArray{T,N,AS.Global}, x)
@@ -359,11 +359,11 @@ Adapt.adapt_storage(::Runtime.Adaptor, x::ROCArray{T,N}) where {T,N} =
 # considered GPU-compatible.
 
 Adapt.adapt_storage(::Type{ROCArray}, xs::AT) where {AT<:AbstractArray} =
-  isbitstype(AT) ? xs : convert(ROCArray, xs)
+    isbitstype(AT) ? xs : convert(ROCArray, xs)
 
 # if an element type is specified, convert to it
 Adapt.adapt_storage(::Type{<:ROCArray{T}}, xs::AT) where {T, AT<:AbstractArray} =
-  isbitstype(AT) ? xs : convert(ROCArray{T}, xs)
+    isbitstype(AT) ? xs : convert(ROCArray{T}, xs)
 
 Adapt.adapt_storage(::Type{Array}, xs::ROCArray) = convert(Array, xs)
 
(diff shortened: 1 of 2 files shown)


src/array.jl
```diff
@@ -519,3 +519,37 @@ end
...
+      maxsize
+    end
+    new_buf = Mem.alloc(A.buf.device, bufsize)
+    copy_size = min(length(A), n) * sizeof(T)
+    if copy_size > 0
+      Mem.transfer!(new_buf, A.buf, copy_size)
+    end
+    Mem.retain(new_buf)
+    if Mem.release(A.buf)
+      Mem.free(A.buf)
```
jpsamaroo, line 549: Missed indentation here

src/array.jl
```diff
@@ -519,3 +519,37 @@ end
...
+    bufsize = if Base.isbitsunion(T)
+      # type tag array past the data
+      maxsize + n
+    else
+      maxsize
+    end
+    new_buf = Mem.alloc(A.buf.device, bufsize)
+    copy_size = min(length(A), n) * sizeof(T)
+    if copy_size > 0
+      Mem.transfer!(new_buf, A.buf, copy_size)
```
jpsamaroo, line 545: Indentation

test/rocarray/base.jl
```diff
@@ -27,6 +27,31 @@ end
...
 end
 
+@testset "resize!" begin
+    a_h = Array(range(1, 10))
+    a_d = a_h |> roc
+    # Resize up
+    resize!(a_h, 15)
+    resize!(a_d, 15)
+    a_h[10:15] .= 15
+    a_d[10:15] .= 15
```
jpsamaroo, line 37: I think you meant to check that `a_h[1:10] .== Array(a_d[1:10])`? The appended bytes are uninitialized, so we can't predict their values.

test/rocarray/base.jl
```diff
@@ -27,6 +27,31 @@ end
...
+    @allowscalar begin
+        @test a_h == a_d
```
jpsamaroo, line 39: This won't work due to uninitialized data, you can just remove this test.

test/rocarray/base.jl
```diff
@@ -27,6 +27,31 @@ end
...
+    end
+    # Keep the size as is
+    resize!(a_h, 15)
+    resize!(a_d, 15)
+    @allowscalar begin
+        @test a_h == a_d
```
jpsamaroo, line 45: As above, test `a_h[1:10] .== Array(a_d[1:10])`

test/rocarray/base.jl
```diff
@@ -27,6 +27,31 @@ end
...
+    end
+    # Resize down
+    resize!(a_h, 3)
+    resize!(a_d, 3)
+    @allowscalar begin
+        @test a_h == a_d
```
jpsamaroo, line 51: As above, test `a_h[1:10] .== Array(a_d[1:10])`

test/rocarray/base.jl
```diff
@@ -27,6 +27,31 @@ end
     @test collect(c) == 3:4
 end
 
+@testset "resize!" begin
```
jpsamaroo, line 30: Can you also add a check, for each of the 3 resize kinds, that `length(a_h) == length(a_d)`?

jpsamaroo (commented):
Great stuff!

test/rocarray/base.jl
matinraayai: I wanted to write them with the same value to check if we can write to it, and to initialize those new bytes to have predictable values.

test/rocarray/base.jl
matinraayai: Even though normal equality between abstract arrays will check for that, I will write a line to check for the length too.

test/rocarray/base.jl
matinraayai: See above comment. I initialize the new elements to have the same value on both host and device. Therefore the equality should pass.

test/rocarray/base.jl
matinraayai: Since both arrays will have the length 3, there is no need for slicing.

test/rocarray/base.jl
jpsamaroo: Right, I misread this, sorry!

test/rocarray/base.jl
jpsamaroo: Nevermind!

jpsamaroo (approved):
LGTM! Just waiting on CI.

jpsamaroo:
Thanks!

PR #353, opened 2023-01-19
Fix `unsafe_free!` not actually freeing

jpsamaroo:
Adds liveness tracking for each HSA/HIP pointer, and allows `unsafe_free!` to mark a pointer as dead; standard buffer refcounting will use this liveness marker to determine if the pointer should be freed.

Also fixes a refcounting off-by-one bug, excessive retains in derived array construction, and moves HSA refcount increment/decrement into memory allocation/free methods.

@pxl-th

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -226,7 +226,7 @@ function Base.unsafe_wrap(::Type{<:ROCArray}, ptr::Ptr{T}, dims::NTuple{N,<:Inte
     @assert isbitstype(T) "Cannot wrap a non-bitstype pointer as a ROCArray"
     sz = prod(dims) * sizeof(T)
     device_ptr = lock ? Mem.lock(ptr, sz, device) : ptr
-    buf = Mem.Buffer(device_ptr, ptr, device_ptr, sz, device, false, false)
+    buf = Mem.Buffer(device_ptr, Ptr{Cvoid}(ptr), device_ptr, sz, device, false, false)
     return ROCArray{T, N}(buf, dims; own=false)
 end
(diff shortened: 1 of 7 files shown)


pxl-th:
@jpsamaroo, as of now, `own` field in `ROCArray` is not used and we always call `Mem.retain` in `ROCArray` ctor.

I think we should not `Mem.retain` a buffer if the array does not `own` it?
It makes sense that things like `reshape(x, ...)`, `@view(x)` do not increase the refcounter and it's the user's responsibility to keep `x` alive?
And on the positive side, we'll get slightly faster finalization of non-owning arrays.

pxl-th:
> @jpsamaroo, as of now, `own` field in `ROCArray` is not used and we always call `Mem.retain` in `ROCArray` ctor.
> 
> I think we should not `Mem.retain` a buffer if the array does not `own` it? It makes sense that things like `reshape(x, ...)`, `@view(x)` do not increase the refcounter and it's the user's responsibility to keep `x` alive? And on the positive side, we'll get slightly faster finalization of non-owning arrays.

Just checked that CUDA.jl does refcount views so maybe we should be consistent... But then the `own` field is not used.

jpsamaroo:
`own` is really supposed to be for `ROCArray`s constructed from arbitrary pointers via `unsafe_wrap`; views and such only need the refcounting to ensure that memory isn't accidentally freed from under the user when they only keep the view and not the original array. But since the `Buffer` already holds information on where the memory comes from (the `host_ptr` fields indicates that a buffer is from `unsafe_wrap`), we can just drop `own` and rely on the existing code.

jpsamaroo:
Latest commits look good to me! Feel free to merge whenever you're happy with this.

pxl-th:
I've dropped `own` field then.

PR #355, opened 2023-01-22
Don't mark `@roc` kernel arguments by default

pxl-th:
I propose we don't mark `@roc` kernel arguments by default.
Consider this example:
```julia
julia> x = [AMDGPU.ones(16) for _ in 1:16];
```

With `mark=true` each element will have `ROCKernelSignal` in its `syncstate` thus keeping `ROCSignal` & `ROCExecutable` with it.
Meaning, we won't be able to free them with `GC.gc()`, not until these elements are used somewhere that calls `wait!` on them.

But every `gpu_call` waits on the kernel, meaning once each element is produced we should be able to free its respective `ROCKernelSignal`.

I think we need to mark only when there is multithreading involved and the user does not want to handle synchronization manually, which is potentially dangerous...

That said, hip calls (rocBLAS) are async, I presume, and thus we need to mark them?

All the tests pass with this change.


jpsamaroo:
I don't think we can do this until we have a task-local queue concept like CUDA.jl (which I'm working on implementing locally). Once we have that, then we can provide `mark` as an optional synchronization mechanism, or remove it in favor of something lighter-weight.

Alternatively, we can make `wait(::ROCKernelSignal)` clear finished signals from the syncstate of its arguments. That would be a non-breaking change.

pxl-th:
As far as I understand, `mark` should be used when the user does not want to handle waiting/synchronization manually? Like multithreaded programming or just stacking bunch of operations on variables?

If so, then I think it is kind of uncommon situation? And maybe we should make it opt-in as to make the user explicitly aware of this mechanism.
It is also ambiguous, without user's waiting/synchronization management.
Like imagine one thread launching a kernel with arguments, and another thread is trying to read those arguments.

And from a perspective of a single thread, `GPUArrays.gpu_call` (and other `@roc` kernel launches in AMDGPU.jl) always wait on a kernel, so there is no need to populate `syncstate`.

I've encountered no issues with marking disabled.

---

> Alternatively, we can make `wait(::ROCKernelSignal)` clear finished signals from the syncstate of its arguments

Like passing them to `wait(::ROCKernelSignal)`, since it has no access to them otherwise?
I'd avoid storing them somewhere in the signal/kernel struct...

jpsamaroo:
The purpose of the `mark!` call is to avoid possible read-write and write-write races, especially considering that we need to span the HSA-HIP synchronization gap when calling into HIP-enabled libs. Many of our tests won't see any changes from removing `mark!`, because they have enough latency from compilation to avoid seeing a race, but it can happen (e.g. if you call into rocBLAS with an array and then immediately try to show the array).

Of course, if we were just worried about AMDGPU-launched kernels (and also had a working task-local blocking queue implementation with an available `AMDGPU.synchronize`), then this would be fine. But with our need to call into HIP libs for all sorts of useful things, this just won't work without some kind of tracking of operations launched on HIP streams.

pxl-th:
So we will always wait on BIP stuff, even when launching HSA kernels, but the user is responsible to wait on HSA kernels

pxl-th:
Let's close for now.

PR #356, opened 2023-01-23
Don't sync on default HIP stream every time

pxl-th:
This PR fixes rocRAND & rocFFT synchronization, which allows us to get rid of default HIP stream synchronization every time on `wait!`.


jpsamaroo (approved):
LGTM! Merge at your leisure :slightly_smiling_face:

PR #363, opened 2023-01-24
Simplify math intrinsics

pxl-th:
Instead of math intrinsic going through [1](https://github.com/JuliaGPU/AMDGPU.jl/blob/3a117c0be77954c9d0fc2caf0c2b6c57c2b584f2/src/device/gcn/math.jl#L87) -> [2](https://github.com/JuliaGPU/AMDGPU.jl/blob/3a117c0be77954c9d0fc2caf0c2b6c57c2b584f2/src/device/gcn/helpers.jl#L5) stages, we can directly perform `ccall`.

This consistently results in faster TTFX & lower memory usage.
Also, IMO, simpler to understand.

- before
```julia
julia> @time sin.(x);
  1.656225 seconds (5.95 M allocations: 338.840 MiB, 4.85% gc time)
```
- this PR
```julia
julia> @time sin.(x);
  1.113379 seconds (2.66 M allocations: 173.616 MiB, 4.01% gc time)
```

### TODO

- [x] Update tests that use `MATH_INTRINSICS` in them, since we no longer use it.
- [x] We can do the same for [OCKL intrinsics](https://github.com/JuliaGPU/AMDGPU.jl/blob/jps/dev/src/device/gcn/wavefront.jl) as well and do direct `ccall`.


jpsamaroo (approved):
Awesome stuff!

pxl-th:
I've updated intrinsics for wavefront as well to perform direct ccall.
And just to note that `GCNIntrinsic` is not used anywhere now.

PR #367, opened 2023-01-29
Pass exception flag through kernel state

jpsamaroo:
We now use GPUCompiler's kernel state support to pass an `AMDGPU.KernelState` object as the first argument to all kernels. This object contains an exception flag, which replaces the exception that we would place in a global variable, which allows us to know exactly which kernels threw an exception (instead of making it appear like all kernels of a given type may have thrown an exception).

One unfortunate side effect of adding this new argument is that it disables kernarg buffer sharing across otherwise identical kernel invocations, because now each `KernelState` object has a unique exception flag. Maybe some kind of dedicated buffer pooling would help here? That would also decrease the newly-added cost of flag allocation in the launch path.

Todo:
- [ ] Pool exception flag buffers
- [ ] Investigate if we can keep kernarg sharing benefits


pxl-th:
Implemented in #423.

PR #370, opened 2023-01-31
Add more MIOpen functionality

pxl-th:
- Max/mean pooling.
- Simplify convolutions.
- Softmax.
- Activations.
- Batch Normalization.
- `mark!`/`wait!` on arguments when calling MIOpen functions.


pxl-th:
Things like dropout *require* passing workspace buffer that was used in forward pass to the backward pass.
Same for batch normalization, which allows passing computed mini-batch mean & inverse variance to the backward pass to speed-up the computation.

From NNlib/Flux we can pass them in rrules and drop during regular calls.

```julia
function rrule()
    y, workspace = MIOpen.dropout(x)
    pullback(dy) = MIOpen.∇dropout(dy, y, x, workspace)
    y, pullback
end
```

src/dnn/activations.jl
```diff
@@ -0,0 +1,142 @@
...
+    x::ROCArray{T}, desc::ActivationDescriptor,
+) where T <: MIOPENFloat
+    y = similar(x)
+    xdesc, ydesc = TensorDescriptor.((x, y))
+    AMDGPU.wait!(x)
+    miopenActivationForward(
+        handle(), desc.handle, Ref{Float32}(1f0), xdesc.handle, x,
+        Ref{Float32}(0f0), ydesc.handle, y) |> check
+    AMDGPU.mark!(y, C_NULL)
+    AMDGPU.wait!(y)
```
jpsamaroo, line 125: Why are you including this `wait!` call here and below?

src/dnn/batchnorm.jl
```diff
@@ -0,0 +1,109 @@
...
+    miopenBatchNormalizationForwardInference(
+        handle(), mode, Ref{Float32}(1f0), Ref{Float32}(0f0),
+        xdesc.handle, x, ydesc.handle, y, bndesc.handle,
+        γ, β, μ, ν, ϵ) |> check
+    AMDGPU.mark!(y, C_NULL)
+    AMDGPU.wait!(y)
+    y
+end
+
+function derive_betta_gamma_descriptors(
```
jpsamaroo, line 86:
```suggestion
function derive_beta_gamma_descriptors(
```

Typo?

jpsamaroo (approved):
LGTM, aside from the double-wait

src/dnn/batchnorm.jl
pxl-th: Yes, thanks!

src/dnn/activations.jl
pxl-th: I guess I was overly cautious. Removed them in all places.

jpsamaroo: Yeah, if everything using `wait!` and `mark!` correctly, then you only need one pair.

PR #374, opened 2023-02-05
Switch to task-focused synchronization model

jpsamaroo:
This PR switches AMDGPU to a task-focused synchronization model, similar to CUDA.jl. In such a model, tasks have their own device and queue assigned, and all GPU operations on that task are launched in serial. That is to say, kernels don't start until any previously-launched kernels have completed. It is thus no longer necessary to `wait` on the result of `@roc`, and instead, one can call `AMDGPU.synchronize()` on the host to wait for all currently-executing kernels (launched from this task) to complete.

There are many reasons why we'd want to switch to this model:
- The rest of the ecosystem generally expects this, as CUDA.jl pioneered this approach
- HSA queues are actually in-order, so there's no reason to pretend that they can execute work concurrently
- It simplifies most common forms of GPU programming, as users are used to thinking about serial computations
- The programming model approximately matches how Julia tasks execute

The existing `wait`-focused model is still fully available, so for the most part, behavior remains the same (assuming users were previously only using one task to launch GPU kernels); the previous single-queue behavior can be enabled by calling `AMDGPU.Runtime.set_alloc_queue_pool_max!([1,0,0])`, which shares a single queue per device, and disables the queue pooling logic for low- and high-priority queues.

Todo:
- [ ] Do the same pooling for HIP streams and track those in TLS
- [ ] Cache task-local ROCm library handles in TLS
- [ ] Track and reuse previously-allocated queue/stream when switching device
- [ ] Add tests
- [ ] Add docs


src/highlevel.jl
```diff
@@ -165,14 +190,34 @@ barrier_or!(queue::ROCQueue, signals::Vector{ROCSignal}) =
...
+end
+
+"""
+    synchronize()
+    synchronize(queue::ROCQueue)
+
+Blocks until all kernels currently executing on `queue` have completed. When
+`queue` is unspecified, the default queue is used.
+"""
+function synchronize(queue::ROCQueue=default_queue())
```
vchuravy, line 211: The one thing that is a tad bit annoying is the name conflicts. But I guess we have to live with that

Noticed that in https://github.com/JuliaGPU/KernelAbstractions.jl/pull/317

src/tls.jl
```diff
@@ -0,0 +1,100 @@
...
+end
+
+"""
+    task_local_state() -> TaskLocalState
+
+Returns the task-local state in the form of a `TaskLocalState`. Automatically
+picks a device and queue if they haven't already been selected.
+"""
+task_local_state() =
+    get!(()->TaskLocalState(), task_local_storage(), :__amdgpu_task_state)
```
vchuravy, line 41:
```suggestion
    get!(()->TaskLocalState(), task_local_storage(), :__amdgpu_task_state)::TaskLocalState
```

I think this is necessary?

vchuravy (commented):
Love it.

luraess:
That's awesome @jpsamaroo ! Thanks, we will look into this with @omlins and @utkinis to finalise AMDGPU support in ImplicitGlobalGrid.jl and ParallelStencil.jl.

src/tls.jl
```diff
@@ -0,0 +1,172 @@
...
+        else
+            @assert queue.priority == priority
+        end
+    end
+    if context === nothing
+        context = HIPContext(device_id(device))
+    end
+    HIP.context!(context)
+    if stream === nothing
+        stream = HIPStream()
```
pxl-th, line 32: Otherwise, querying for default device in `HIPStream()` causes stackoverflow.

```suggestion
        stream = HIPStream(device)
```

src/hip/HIP.jl
```diff
@@ -108,16 +101,50 @@ function HIPStream(device::HIPDevice; priority=:normal)
...
+    priority == :normal && return Cint(0)
+    priority == :high && return Cint(-1)
+    priority == :low && return Cint(1)
+    throw(ArgumentError("""
+    Invalid HIP priority symbol: $priority.
+    Valid values are: `:normal`, `:low`, `:high`.
+    """))
+end
+
+function hipStreamGetPriority(stream::hipStream_t)
```
jpsamaroo, line 144: Maybe just call this `priority(stream)`?

src/hip/HIP.jl
```diff
@@ -90,15 +90,8 @@ mutable struct HIPStream
     device::HIPDevice
 end
 function HIPStream(device::HIPDevice; priority=:normal)
```
jpsamaroo, line 92: I wonder if we should just get the device from the active context?

src/runtime/queue.jl
```diff
@@ -161,6 +175,7 @@ function ROCQueue(device::ROCDevice; priority::Symbol=:normal, pooled::Bool=fals
     @log_finish(:alloc_queue, (;alloc_id), (;queue=reinterpret(UInt64, queue.queue)))
     return queue
 end
+ROCQueue(; kwargs...) = ROCQueue(get_default_device(); kwargs...)
```
pxl-th: Duplicates one above
```suggestion
```

jpsamaroo: Yeah, pushed a fixup

PR #375, opened 2023-02-06
Store `HSA.Signal` instead of `Ref{HSA.Signal}` in `ROCSignal`

pxl-th:
We always unref `Ref{HSA.Signal}` before using it anywhere, so instead we can unref it right after the creation, which will save us from unnecessary allocations later on.


jpsamaroo:
Thanks!

PR #376, opened 2023-02-07
Reduce allocations

pxl-th:
- Remove runtime dispatch in `getinfo` calls.
- Store `HSA.Executable` instead of `Ref{HSA.Executable}` in `ROCExecutable`.

Before:
```julia
0.262667 seconds (305.51 k allocations: 14.300 MiB)
3.865236 seconds (1.08 M allocations: 33.484 MiB, 0.37% gc time)
```
This PR:
```julia
5.816722 seconds (8.27 M allocations: 543.529 MiB, 2.75% gc time, 5.58% compilation time)
3.420838 seconds (556.49 k allocations: 19.608 MiB, 0.60% gc time)
```


jpsamaroo (approved):
Awesome!

PR #383, opened 2023-02-14
ROCSignal: Disable IPC by default

jpsamaroo:
Because it's about 300X more expensive and probably uses more resources


michel2323:
Just want to add for reference that one some GPUs the slowdown is actually 10000x:

Issue #372, opened 2023-02-02
AMDGPU.jl master is broken on Julia 1.7


utkarsh530:
Hi, I am trying to get `AMDGPU.jl` to work with the mentioned GPU. Although I can create `ROCArray`, things like kernel launches and `A\b` throws an error. The `] test AMDGPU` doesn't pass as well. 

Code + Stacktrace: https://pastebin.com/Rtt1n534
```
julia> AMDGPU.versioninfo()
HSA Runtime (ready)
- Path: /home/utkarsh/.julia/artifacts/169e9a933365dbab116a56e28d5e468c5c7a10c5/lib/libhsa-runtime64.so
- Version: 1.1.0
ld.lld (ready)
- Path: /home/utkarsh/.julia/artifacts/836082a6a472e357ce62db92868086df6735be27/bin/lld
ROCm-Device-Libs (ready)
- Path: /home/utkarsh/.julia/artifacts/9fdb94cae0fc41f8cb5cf92a1d81e1edf8a71e98/amdgcn/bitcode
HIP Runtime (ready)
- Path: /home/utkarsh/.julia/artifacts/858cd7ded8156b94aae719650397d091bd78c4dc/hip/lib/libamdhip64.so
rocBLAS (ready)
- Path: /home/utkarsh/.julia/artifacts/c7f4295c8fc1b930aa2d18ed575540df6c9de02e/rocblas/lib/librocblas.so
rocSOLVER (MISSING)
rocALUTION (MISSING)
rocSPARSE (ready)
- Path: /home/utkarsh/.julia/artifacts/4db863b21264fac571033dbd49ff6efc3f393cbf/rocsparse/lib/librocsparse.so
rocRAND (ready)
- Path: /home/utkarsh/.julia/artifacts/3c565b6e353eb2ec9345ac55ba0f2d6be3444bea/hiprand/lib/../../rocrand/lib/librocrand.so.1
rocFFT (MISSING)
MIOpen (MISSING)
HSA Agents (2):
- CPU-XX [AMD FX(tm)-8350 Eight-Core Processor]
- GPU-0213f2e93ef02184 [Radeon RX Vega (gfx900)]
```

```
julia> versioninfo()
Julia Version 1.7.3
Commit 742b9abb4d (2022-05-06 12:58 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: AMD FX(tm)-8350 Eight-Core Processor
[...]

PR #392, opened 2023-03-17
Add blas wrappers for triangular matrix mul / div

pxl-th:
We had method definitions for:
`Transpose{T,UpperTriangular{T,ROCMatrix{T}}}`
But not for:
`LowerTriangular{T, Transpose{T, ROCMatrix{T}}}`

Which is what you get from:
```julia
julia> x = ROCArray(triu(rand(Float32, 4, 4)));

julia> x = transpose(UpperTriangular(x))
4×4 LowerTriangular{Float32, Transpose{Float32, ROCMatrix{Float32}}}:
 0.939737    ⋅         ⋅         ⋅ 
 0.0391901  0.184103   ⋅         ⋅ 
 0.578203   0.059085  0.419264   ⋅ 
 0.245061   0.34397   0.923544  0.710086
```

This PR adds those definitions for BLAS 2 & 3 as well as tests.

### DIFF
--- a/src/blas/highlevel.jl
+++ b/src/blas/highlevel.jl
@@ -64,6 +64,44 @@ Base.argmax(xs::ROCBLASArray{<:ROCBLASReal}) = iamax(xs)
 #
 ############
 
+for (t, uploc, isunitc) in (
+    (:LowerTriangular, 'L', 'N'),
+    (:UnitLowerTriangular, 'L', 'U'),
+    (:UpperTriangular, 'U', 'N'),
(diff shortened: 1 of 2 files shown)


pxl-th:
For generic `A \ B` we'd need rocSOLVER and something similar to [this](https://github.com/JuliaGPU/CUDA.jl/blob/f3156e3db4394822daa5225fac21e15e9c5a77de/lib/cusolver/linalg.jl#L30), which would solve #372.

PR #395, opened 2023-03-18
Reduce memory usage

pxl-th:
- Specify types for global cache in memoize macro.
- Memoize device regions and pools.
- Use in-place `filter!` in some places where it is safe to do so.
- Use raw HSA signal in HostCall instead of `ROCSignal`. `ROCSignal` was finalized too early leading to periodic crashes when exiting Julia.


jpsamaroo:
I'm pretty sure that the Distributed implementation sped up the tests significantly for me, obviously at the expense of more memory usage. Are you sure that all the tests ran correctly?

pxl-th:
> I'm pretty sure that the Distributed implementation sped up the tests significantly for me, obviously at the expense of more memory usage. Are you sure that all the tests ran correctly?

My bad, the speedup was indeed substantial. With distributed tests take ~20 min, without ~55 min.
Noticeably, `GPUArrays - broadcast` is the longest of them, taking almost 10 minutes.

PR #407, opened 2023-04-04
Update to GPUCompiler 0.19 & LLVM 5

pxl-th:
- [ ] Use `@dispose` for builders?


pxl-th:
Do you have timespan logging enabled?

pxl-th:
It happens, because we've moved from FunctionSpec to MethodInstance, but I forgot to update the logging stuff.
Will be fixed shortly

matinraayai:
@pxl-th it works now thanks.

PR #416, opened 2023-04-25
Use GPUArray storage handling

pxl-th:
Ref https://github.com/JuliaGPU/GPUArrays.jl/pull/468.


pxl-th:
Closed in favor of #481.

PR #417, opened 2023-04-27
Fix host synchronization

pxl-th:
- Fix typo in `isempty` for list of active kernels.
- Move to the next kernel in the list only when done waiting on the previous kernel.
Previously this may have led to a following sync issue:
```
- monitor pops kernel to wait on, moves to the next kernel and starts waiting on the pop'd kernel
- host calls AMDGPU.synchronize(), but since it is empty it returns immediately
- monitor finishes waiting on pop'd kernel
```
- Remove scratch alloc limiter.

### DIFF
--- a/src/runtime/kernel.jl
+++ b/src/runtime/kernel.jl
@@ -119,14 +119,16 @@ function ROCKernel(kernel #= ::HostKernel =#; localmem::Int=0)
     group_segment_size = UInt32(group_segment_size + localmem)
     private_segment_size = executable_symbol_kernel_private_segment_size(exec_symbol)
     if private_segment_size > MAXIMUM_SCRATCH_ALLOCATION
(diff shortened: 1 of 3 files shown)


src/runtime/kernel.jl
```diff
@@ -119,14 +119,16 @@ function ROCKernel(kernel #= ::HostKernel =#; localmem::Int=0)
     group_segment_size = UInt32(group_segment_size + localmem)
     private_segment_size = executable_symbol_kernel_private_segment_size(exec_symbol)
```
vchuravy, line 120: Is this advisory? E.g. does the kernel run correctly if it get's less than the requested memory?

pxl-th: Probably not... If decreased to a too low value, kernel crashes.
I remember it was added baaack when we were dealing with `HSA_STATUS_ERROR_OUT_OF_RESOURCES` queue errors.
But since then there's been a lot of fixes regarding memory usage, so maybe worth removing it.

pxl-th: Removed. Everything works fine without it.

PR #420, opened 2023-05-01
Add device selection in current task by ID

luraess:
This PR adds `AMDGPU.device_id!` to allow switching device in the current task. It also extends the doc.


pxl-th (approved):

pxl-th:
Thanks!

PR #419, opened 2023-04-30
Skip host synchronization when it is safe to do so

pxl-th:
Closes #405.

Since we use HSA for Julia kernels & HIP for other library calls (gemm, etc.) we have to perform wait at `hsa | hip` boundary, for example:
`(x .* x) *  x `
`  hsa    | hip`

However, when using only hsa or only hip kernels in a row we can rely on on-device serialization and skip host wait.
This allows us to dispatch kernels asynchronously.

The only restriction is that you have to use the same HSA queue or HIP stream.
Which is fine, since we've moved to TLS.

Changes to the `SyncState` are the following:
- Maintain `same_queue` & `same_stream` to check if all signals or streams belong to the same queue / are equal.
- If either `same_queue == false` or `same_stream == false` fallback to the old wait behavior.
- Otherwise, if `SyncState` contains only HSA signals and we are dispatching another HSA kernel, skip `wait!` if it is called.
- To signal that you are dispatching another HSA kernel or doing another HIP call and want to skip (potentially) host wait, use `hsa_wait!` or `hip_wait!`.

`hsa_wait!` always waits for any HSA signal if it is present in a `SyncState`. It is meant to be used right before HIP library call, e.g. before `gemm`.
`hip_wait!` always waits for any HIP stream if it is present in a `SyncState`. It is meant to be used right before HSA kernel dispatches, e.g. inside `@roc` macro.

- When skipping host wait, for example for HSA, remove all HSA signals from `SyncState` except the last one. This is to ensure we synchronize if the next OP is HIP library call.
- Avoid duplication in SyncState. Code like `broadcast!(cos, x, x)` previously would push same signal twice into `x`'s `SyncState`.

- Synchronize on `HIPEvent` instead of `HIPStream` for HIP-based libraries.
  `HIPEvent` is created at the moment of `mark!`.

### Code

```julia
using BenchmarkTools
using AMDGPU

x = ROCArray(rand(Float32, 128, 128))

function matmul(x)
    t1 = x * x
    t2 = x * t1
    t3 = x * t2
    AMDGPU.synchronize()
    AMDGPU.unsafe_free!(t1)
    AMDGPU.unsafe_free!(t2)
    AMDGPU.unsafe_free!(t3)
    return
end

function el_mul(x)
    t1 = x .* x
    t2 = x .* t1
    t3 = x .* t2
    AMDGPU.synchronize()
    AMDGPU.unsafe_free!(t1)
    AMDGPU.unsafe_free!(t2)
[...]

PR #298, opened 2022-09-26
Rocsparse Support for AMDGPU.jl

matinraayai:
@jpsamaroo this is my somewhat working branch of the rocsparse port. Please see if you are able to debug the ```broadcast.jl``` and ```linalg.jl``` tests in the rocsparse folder.

PR #422, opened 2023-05-04
Migrate to stream-ordered allocator for arrays

### LINKED ISSUE #418: rocBLAS synchronization issue?
Calling this code several times in a row (`tt()` function):
```julia
function _step(X, W, H)
    X * (W * H)
end

function tt()
    nrow = 1024 * 20
    X = Base.rand(Float32, nrow, nrow)
    W = Base.rand(Float32, nrow, nrow)
    H = Base.rand(Float32, nrow, nrow)

    RX = ROCArray(X)
    RW = ROCArray(W)
    RH = ROCArray(H)
    gpu_res = _step(RX, RW, RH)
    return
end
```

For some reason results in this error:
```julia
┌ Error: Memory Fault on GPU-XX [AMD Radeon RX 6700 XT (gfx1030)] at 0x00007f628aea4000:
│ PAGE_NOT_PRESENT, READ_ONLY, NX, HOST_ONLY, DRAMECC, IMPRECISE, SRAMECC, HANG
│ GPU kernels are now hung, please restart Julia
└ @ AMDGPU.Runtime ~/.julia/dev/AMDGPU/src/runtime/fault.jl:25
```

Even though we always synchronize on arguments before calling next rocBLAS gemm.

pxl-th:
TODO: add detailed PR description

Builds upon #419 
Fixes #418

### DIFF
--- a/src/AMDGPU.jl
+++ b/src/AMDGPU.jl
@@ -78,7 +78,24 @@ module Runtime
     include(joinpath("runtime", "signal.jl"))
     include(joinpath("runtime", "dims.jl"))
     module Mem
-        include(joinpath("runtime", "memory.jl"))
+        import AMDGPU
+        import AMDGPU: HSA
+        if AMDGPU.hip_configured
+            import AMDGPU: HIP
+        end
+        import AMDGPU: Runtime
+        import .Runtime: ROCDevice, ROCSignal, ROCMemoryRegion, ROCMemoryPool, ROCDim, ROCDim3
+        import .Runtime: DEVICES, check, get_region, get_memory_pool, get_handle
+
+        using Preferences
+
+        abstract type AbstractAMDBuffer end
+
+        include(joinpath("runtime", "memory", "utils.jl"))
+        include(joinpath("runtime", "memory", "memory.jl"))
+        include(joinpath("runtime", "memory", "pooled.jl"))
+        include(joinpath("runtime", "memory", "hip.jl"))
+        include(joinpath("runtime", "memory", "refcount.jl"))
     end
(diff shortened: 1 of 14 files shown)


mzy2240:
I am wondering if this PR would also fix the synchronization issue in https://github.com/JuliaGPU/AMDGPU.jl/pull/298

pxl-th:
In general this PR fixes use-after-free issues, where previously one could hit a scenario where:
- call a HIP-based kernel (e.g. rocBLAS gemm)
- GC frees arrays you passed as arguments before kernel finishes
- you get use-after-free error and/or a hang

But the issue with HIP-based allocations at the moment is that they are drawn from a HIP memory pool which grows on-demand and you can limit the total size of the pool only starting from ROCm 5.5+ via [`hipDeviceSetLimit`](https://github.com/ROCm-Developer-Tools/hipamd/blob/c1741e9b671a58080b7150e11d9130e6a2f22dcb/src/hip_device_runtime.cpp#L480).

And when the pool grows to near 100% of the total memory available on the device it kills our HSA queue since it cannot allocate necessary resources for a kernel dispatch.

Interestingly, you can [get](https://github.com/ROCm-Developer-Tools/hipamd/blob/474e8620099a463ad2ced821ae7400609b29bf7f/src/hip_device_runtime.cpp#L397) the memory limit of the pool on ROCm 5.4 but not set it.

And constraining pool growth from the Julia side seems unreliable at the moment.

pxl-th:
Superseded by https://github.com/JuliaGPU/AMDGPU.jl/pull/423

PR #433, opened 2023-06-22
Update FFT interface

mfherbst:
Sharing as far as I got with an hour of work. Definitely more fixes needed and up for graps if someone wants to take over.


mfherbst:
`fft(a::ROCArray)` now works without crash, but gives the wrong answer (basically returns uninitalised memory). I had to move around the stream passing and task local storage stuff, which I'm not too sure if this is reasonable. Will do more investigation later.

pxl-th:
Superseded by #443

Issue #436, opened 2023-06-27
Cannot `unsafe_wrap` a device array if `lock=false`


Keluaa:
Hello, I am trying to wrap a GPU array allocated from another library with a `ROCArray`:

```julia
julia> roc_array = Base.unsafe_wrap(ROCArray, pointer(my_array), size(my_array); lock=false)
ERROR: MethodError: no method matching AMDGPU.Runtime.Mem.Buffer(::Ptr{Float64}, ::Ptr{Nothing}, ::Ptr{Float64}, ::Int64, ::ROCDevice, ::Bool, ::Bool)

Closest candidates are:
  AMDGPU.Runtime.Mem.Buffer(::Ptr{Nothing}, ::Ptr{Nothing}, ::Ptr{Nothing}, ::Int64, ::ROCDevice, ::Bool, ::Bool)
   @ AMDGPU /home/briandl/.julia/packages/AMDGPU/ZRqNQ/src/runtime/memory.jl:27

Stacktrace:
 [1] unsafe_wrap(::Type{ROCArray}, ptr::Ptr{Float64}, dims::Tuple{Int64, Int64}; device::ROCDevice, lock::Bool)
   @ AMDGPU /home/briandl/.julia/packages/AMDGPU/ZRqNQ/src/array.jl:228
 [2] top-level scope
   @ REPL[19]:1
```

I pass `lock=false` to `unsafe_wrap` since `AMDGPU.Mem.lock` assumes that the pointer is a host pointer (if I understood the source code correctly).

[But at this line](https://github.com/JuliaGPU/AMDGPU.jl/blob/6028decb42643eb422d773d53ce6d6cd4a337b9c/src/array.jl#L227), if `lock=false`, `ptr` is not converted to a `Ptr{Cvoid}` as expected by `Mem.Buffer`, hence the bug.

jpsamaroo:
Good find! That should be a one line fix, plus maybe a small test, if you wouldn't mind submitting a PR?

PR #438, opened 2023-07-05
Fix `unsafe_wrap` for device arrays

Keluaa:
Here is my draft of a fix for #436.

What I thought would be a simple task left me quite confused about some things.

My initial understanding of the kwarg `lock` of `unsafe_wrap` was:

- `true` (default): copy a CPU array to the GPU and wrap it to a ROCArray, while keeping a reference to the CPU array
- `false`: wrap a GPU array into a ROCArray

But if we set `host_ptr` of the `Mem.Buffer` to the `ptr` given to `unsafe_wrap` (which is wrong if `ptr` is a GPU pointer), then we would have a segfault e.g. in `Mem.transfer!`:
https://github.com/JuliaGPU/AMDGPU.jl/blob/767adb15fbfd691204e6ea969ebb9e1506912132/src/runtime/memory.jl#L647-L650

Therefore, I thought I should add another kwarg `is_device_ptr`, and if `true`, no `host_ptr` should remain `0`.
Something like:
```julia
function Base.unsafe_wrap(
    ::Type{<:ROCArray}, ptr::Ptr{T}, dims::NTuple{N,<:Integer};
    device=device(), lock::Bool=true, is_device_ptr::Bool=false
) where {T,N}
    @assert isbitstype(T) "Cannot wrap a non-bitstype pointer as a ROCArray"
    sz = prod(dims) * sizeof(T)

    if !is_device_ptr
        host_ptr = Ptr{Cvoid}(ptr)
        device_ptr = lock ? Mem.lock(ptr, sz, device) : Ptr{Cvoid}(ptr)
    else
        host_ptr = C_NULL
        device_ptr = Ptr{Cvoid}(ptr)
    end

    buf = Mem.Buffer(device_ptr, host_ptr, device_ptr, sz, device, false, false)
    return ROCArray{T, N}(buf, dims)
end
```

But then the case `lock=true` and `is_device_ptr=false` produces a ROCArray with a buffer with no pointers to the memory of the `device`, and I don't think this makes sense either.

Should we get rid of the `lock` kwarg then?

There is also issues about memory management: the retuned array will gain ownership of the device pointer it is given.
With the test I added, the buffer will be freed twice.
Is there any way to make `Mem.release` or `Mem.free` detect that the array was built with `unsafe_wrap` from an externally allocated array and skip deallocation?

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -224,7 +224,7 @@ end
 function Base.unsafe_wrap(::Type{<:ROCArray}, ptr::Ptr{T}, dims::NTuple{N,<:Integer}; device=device(), lock::Bool=true) where {T,N}
     @assert isbitstype(T) "Cannot wrap a non-bitstype pointer as a ROCArray"
     sz = prod(dims) * sizeof(T)
-    device_ptr = lock ? Mem.lock(ptr, sz, device) : ptr
+    device_ptr = lock ? Mem.lock(ptr, sz, device) : Ptr{Cvoid}(ptr)
     buf = Mem.Buffer(device_ptr, Ptr{Cvoid}(ptr), device_ptr, sz, device, false, false)
     return ROCArray{T, N}(buf, dims)
 end
--- a/src/array.jl
+++ b/src/array.jl
@@ -221,14 +221,25 @@ function Base.copy(X::ROCArray{T}) where T
     Xnew
 end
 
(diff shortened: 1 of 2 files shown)


pxl-th:
Superseded by #442.

It adds better support for pinned host memory and fixes issues with wrapped device memory.
See [this section](https://amdgpu.juliagpu.org/dev/memory/#Wrapping-in-ROCArray) for documentation.

PR #439, opened 2023-07-06
Fix docs for HIP

luraess:
Fix and homogenise naming convention for using AMDGPU with HIP backend for kernel launch. Remove the statement about combat with CUDA naming which is not working.


luraess:
Wondering if this is stil valid statement https://amdgpu.juliagpu.org/dev/#Navi-2-(GFX103x)-support

pxl-th (approved):

PR #471, opened 2023-08-16
`AMDGPU.@elapsed`

carstenbauer:
This PR adds the functionality to measure the elapsed time between two hip events. Inspired by CUDA.jl, it exposes a macro `AMDGPU.@elapsed` for end users.

I tested the new macro in a simple memory bandwidth benchmark (just copying/transferring memory) and it seems to work. That is, the resulting memory bandwidth is reasonable. ~However, what I'm unsure about is the line:~

```julia
t0, t1 = HIPEvent(C_NULL), HIPEvent(C_NULL)
```

~I chose `C_NULL` for the stream here (which according to [this](https://rocm-developer-tools.github.io/HIP/group__Event.html#gad4128b815cb475c8e13c7e66ff6250b7) waits for all commands on all streams to finish). But maybe that should be `AMDGPU.stream()` or something else?~

UPDATE: We use the task-local `stream()` instead of `C_NULL` now.

(cc @jpsamaroo, @luraess)

### DIFF
--- a/src/highlevel.jl
+++ b/src/highlevel.jl
@@ -307,3 +307,22 @@ function launch_configuration(
 )
     HIP.launch_configuration(kern.fun; shmem, max_block_size)
 end
+
+"""
+    @elapsed ex
+
+A macro to evaluate an expression, discarding the resulting value, instead returning the
+number of seconds it took to execute on the GPU, as a floating-point number.
+"""
+macro elapsed(ex)
+    quote
+        current_stream = stream()
+        t0 = HIP.HIPEvent(current_stream; do_record=false, disable_timing=false)
+        t1 = HIP.HIPEvent(current_stream; do_record=false, disable_timing=false)
+        HIP.record(t0)
+        $(esc(ex))
+        HIP.record(t1)
+        HIP.synchronize(t1)
+        HIP.elapsed(t0, t1)
+    end
+end
--- a/src/highlevel.jl
+++ b/src/highlevel.jl
@@ -317,8 +317,8 @@ number of seconds it took to execute on the GPU, as a floating-point number.
 macro elapsed(ex)
     quote
         current_stream = stream()
-        t0 = HIP.HIPEvent(current_stream; do_record=false, disable_timing=false)
-        t1 = HIP.HIPEvent(current_stream; do_record=false, disable_timing=false)
+        t0 = HIP.HIPEvent(current_stream; do_record=false, timing=true)
+        t1 = HIP.HIPEvent(current_stream; do_record=false, timing=true)
         HIP.record(t0)
         $(esc(ex))
         HIP.record(t1)
--- a/src/hip/event.jl
+++ b/src/hip/event.jl
@@ -62,3 +62,31 @@ function HIPEvent(stream::hipStream_t; do_record::Bool = true)
     event
 end
 HIPEvent(stream::HIPStream; do_record::Bool = true) = HIPEvent(stream.stream; do_record)
+
+"""
+    elapsed(start::HIPEvent, stop::HIPEvent)
+
+Computes the elapsed time between two events (in seconds).
+"""
+function elapsed(start::HIPEvent, stop::HIPEvent)
+    time_ref = Ref{Cfloat}()
+    hipEventElapsedTime(time_ref, start, stop)
+    return time_ref[]/1000
+end
+
+"""
+    @elapsed ex
+
+A macro to evaluate an expression, discarding the resulting value, instead returning the
+number of seconds it took to execute on the GPU, as a floating-point number.
+"""
+macro elapsed(ex)
+    quote
+        t0, t1 = HIPEvent(C_NULL), HIPEvent(C_NULL)
+        record(t0)
+        $(esc(ex))
+        record(t1)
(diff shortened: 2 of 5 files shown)


src/hip/event.jl
```diff
@@ -62,3 +62,31 @@ function HIPEvent(stream::hipStream_t; do_record::Bool = true)
...
+
+"""
+    @elapsed ex
+
+A macro to evaluate an expression, discarding the resulting value, instead returning the
+number of seconds it took to execute on the GPU, as a floating-point number.
+"""
+macro elapsed(ex)
+    quote
+        t0, t1 = HIPEvent(C_NULL), HIPEvent(C_NULL)
```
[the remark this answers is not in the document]
carstenbauer: That's what I wrote above. `C_NULL` is the "safe thing" in the sense that it waits for all streams to finish. But I think you're right, `stream()` is probably the "right thing" :)

carstenbauer:
I'm now using `stream()` (i.e. the task local stream) instead of `C_NULL` (i.e. wait for all streams).

Seems to work:
```julia
julia> using GPUInspector, AMDGPU

julia> memory_bandwidth();
Memory Bandwidth (GiB/s):
 └ max: 1144.27

julia> host2device_bandwidth();
Host <-> Device Bandwidth (GiB/s):
 └ max: 17.47
```

src/highlevel.jl
```diff
@@ -307,3 +307,22 @@ function launch_configuration(
...
+"""
+    @elapsed ex
+
+A macro to evaluate an expression, discarding the resulting value, instead returning the
+number of seconds it took to execute on the GPU, as a floating-point number.
+"""
+macro elapsed(ex)
+    quote
+        current_stream = stream()
+        t0 = HIP.HIPEvent(current_stream; do_record=false, disable_timing=false)
```
vchuravy, line 320: Why are you doing `disable_timing=false` I would expect that you want the increased accuracy.

src/hip/event.jl
```diff
@@ -50,9 +50,13 @@ function synchronize(event::HIPEvent)
     return
 end
 
-function HIPEvent(stream::hipStream_t; do_record::Bool = true)
+function HIPEvent(stream::hipStream_t; do_record::Bool = true, disable_timing=true)
```
vchuravy, line 53:
```suggestion
function HIPEvent(stream::hipStream_t; do_record::Bool = true, timing=false)
```

src/hip/event.jl
```diff
@@ -50,9 +50,13 @@ function synchronize(event::HIPEvent)
...
     event_ref = Ref{hipEvent_t}()
-    hipEventCreateWithFlags(event_ref, hipEventDisableTiming) |> check
+    if disable_timing
```
vchuravy, line 55:
```suggestion
    if !timing
```

src/hip/event.jl
jpsamaroo:
```suggestion
function HIPEvent(stream::hipStream_t; do_record::Bool = true, timing::Bool = true)
```

src/hip/event.jl
```diff
@@ -61,4 +65,15 @@ function HIPEvent(stream::hipStream_t; do_record::Bool = true)
     end
     event
 end
-HIPEvent(stream::HIPStream; do_record::Bool = true) = HIPEvent(stream.stream; do_record)
+HIPEvent(stream::HIPStream; kwargs...) = HIPEvent(stream.stream; kwargs...)
+
+"""
+    elapsed(start::HIPEvent, stop::HIPEvent)
+
+Computes the elapsed time between two events (in seconds).
```
jpsamaroo, line 73:
```suggestion
Computes the elapsed time between two events (in seconds).

See also [`@elapsed`](@ref).
```

jpsamaroo (approved):
Looks great! Just nitpicks from my side, and of course we want a test or two.

src/highlevel.jl
```diff
@@ -307,3 +307,22 @@ function launch_configuration(
...
+        t1 = HIP.HIPEvent(current_stream; do_record=false, disable_timing=false)
+        HIP.record(t0)
+        $(esc(ex))
```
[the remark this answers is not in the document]
carstenbauer: Seems unnecessary to me (and CUDA.jl also doesn't do it). But it's your call :)

src/hip/event.jl
carstenbauer: @jpsamaroo you want to change the default to `timing=true`? The behavior before the PR corresponds to `timing=false`.

src/hip/event.jl
```diff
@@ -61,4 +65,17 @@ function HIPEvent(stream::hipStream_t; do_record::Bool = true)
...
+
+See also [`@elapsed`](@ref).
+"""
+function elapsed(start::HIPEvent, stop::HIPEvent)
+    time_ref = Ref{Cfloat}()
+    hipEventElapsedTime(time_ref, start, stop)
```
pxl-th, line 79:
```suggestion
    hipEventElapsedTime(time_ref, start, stop) |> check
```

pxl-th (approved):

pxl-th:
Thanks!

PR #487, opened 2023-09-06
Remove default device stuff

pxl-th:
Remove default device to simplify things.
Now the default device is the first device in the list returned by `AMDGPU.devices()`.
And when changing the device with `AMDGPU.device!` or `AMDGPU.device_id!`, IT becomes the default device for TLS.

CC @luraess.


luraess:
Some basic test with multi-GPU on multi-node selection on LUMI seems to work fine using `device`.

luraess:
Thanks 🙏 !

PR #497, opened 2023-09-19
[rocSOLVER] Add a method for geqrf!

amontoison:
We don't want to allocate the vector `tau` everytime that we perform a (reduced) QR decomposition.


amontoison:
@pxl-th
Is it possible to regenerate `libcusolver.jl` to include the wrappers for `orgqr`?

pxl-th:
> Is it possible to regenerate libcusolver.jl to include the wrappers for orgqr!?

I think last time I tried Clang.jl to generate wrappers for rocSOLVER it was giving me an error, should've opened respective issue.
Therefore I wrote part of them manually.

Will try again today

PR #503, opened 2023-09-29
Enable 5.4 JLLs on LLVM <16

jpsamaroo:
Now that we have working ROCm 5.4 JLLs for the core libs, we should be able to enable this. Of course, we may want to consider if we need to warn users that rocBLAS et. al are not yet available.


pxl-th (approved):
It would be beneficial to turn them into lazy artifacts at some point.
If people are relying mostly on system-wide installation, then there is no need to download them.

luraess:
> It would be beneficial to turn them into lazy artifacts at some point.
> 
> If people are relying mostly on system-wide installation, then there is no need to download them.

Or do something like CUDA did in latest release?

PR #510, opened 2023-10-05
Default to non-blocking synchronize

luraess:
Transition to non-blocking synchronize behaviour by default in order to enable similar behaviour as in CUDA.jl and to comply with KerneAbstractions approach to perform tasks in a cooperative fashion.


luraess:
Seems no other changes are needed, but could be that in some places for `AMDGPU.synchronize()` was called we do now need to explicitly say it should be blocking (or not).

luraess:
Addressed by #512

PR #511, opened 2023-10-06
Switch tests to ReTestItems.jl

pxl-th:
One caveat for now is that we have to manually specify each test item from GPUArrays testsuite.


luraess:
Cool 🚀

PR #512, opened 2023-10-07
Use non-blocking synchronization by default

pxl-th:
Use non-blocking synchronization by default.
Waiting on a stream is done by launching host function on that stream, that triggers an event once it is processed, similar to how CUDA.jl does it.
This requires ROCm 5.4+, therefore we have a fallback that works on previous versions as well.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -135,3 +135,28 @@ Once all of this is setup properly, you should be able to do `using AMDGPU`
 successfully.
 
 See the [Quick Start](@ref) documentation for an introduction to using AMDGPU.jl.
+
+## Preferences
+
+AMDGPU.jl supports setting
+[preferences](https://github.com/JuliaPackaging/Preferences.jl).
+Template of `LocalPreferences.toml` with all options:
+
+```toml
+[AMDGPU]
+# If `true` then use ROCm libraries provided by artifacts.
+# However, not all ROCm libraries are available as artifacts.
+use_artifacts = false
+# Use mixed-mode ROCm. This will use device libraries from artifacts,
+# but the rest of the ROCm libraries from system-wide installation.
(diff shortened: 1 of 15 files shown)


luraess:
Maybe we could also add a `LocalPreferences.toml` file to provide a template as done in CUDA.jl (https://github.com/JuliaGPU/CUDA.jl/blob/master/LocalPreferences.toml):

`LocalPreferences.toml`:
```
[AMDGPU]
# whether to use a local AMDGPU installation
#use_artifacts = true

# whether to use a nonblocking synchronization mechanism,
# making it possible to do use cooperative multitasking.
#nonblocking_synchronization = true
```

docs/src/index.md
luraess, line 1: Maybe we could add this template `.toml` as file in `src` of AMDGPU @pxl-th ?

src/hip/stream.jl
```diff
@@ -66,22 +69,72 @@ function _low_latency_synchronize(stream::HIPStream)
...
+        # Wait for `event`.
+        Threads.@spawn begin
+            Base.wait(event)
+            close(timer)
+        end
+    end
+    return
+end
+
+function non_blocking_synchronize_old(stream::HIPStream)
```
luraess, line 117:
```suggestion
function nonblocking_synchronize_old(stream::HIPStream)
```
else it fails

pxl-th: Thanks!

docs/src/index.md
pxl-th: I'm hesitant about adding `LocalPreferences.toml` since it is meant to be local on your machine (from the docs):
> allowing for users themselves to override those preferences with their own settings in the LocalPreferences.toml file, which should be .gitignored as the name implies.

I also use it quite heavily for AMDGPU now and having it in the repo would make it tedious.

PR #585, opened 2024-01-31
Expose alpha and beta parameters in LinearAlgebra.mul!

wbernoudy:
Keeping the defaults to alpha = 1 and beta = 0 should mean this does not affect current API usage, but ensures LinearAlgebra.mul! does not fall back to general matrix multiply methods when passing in `ROCMatrix`s.


pxl-th:
Thanks!

PR #605, opened 2024-02-28
Mark ccalls as gc safe

pxl-th:


luraess:
Thanks for that

pxl-th:
I thought it might help with https://github.com/JuliaGPU/AMDGPU.jl/issues/606, but the issue seems to be somewhere else.

PR #630, opened 2024-05-15
Remove MIOpen softmax & batchnorm

pxl-th:
NNlib has pure Julia implementation for both and MIOpen softmax has issues: https://github.com/ROCm/MIOpen/issues/2966


atamazov:
Note that https://github.com/ROCm/MIOpen/issues/2966 won't manifest itself unless the same instance of the library is used to get softmax for different datatypes.

The fix in on the way: https://github.com/ROCm/MIOpen/pull/2992

pxl-th:
> Note that https://github.com/ROCm/MIOpen/issues/2966 won't manifest itself unless the same instance of the library is used to get softmax for different datatypes.

Yes, that's exactly what we were doing. We run tests for different types in the same session and users are generally not restricted to what types they can use.

PR #634, opened 2024-05-17
Allow disabling eager GC

pxl-th:
Either call `AMDGPU.eager_gc!(false)` or set in `LocalPreferences.toml` file following:
```toml
[AMDGPU]
eager_gc = false
```

@luraess


luraess:
Great. Will test!

PR #644, opened 2024-06-14
Implement Base.dataids(A::ROCArray)

ahojukka5:
A similar is implemented in CUDA.jl.

See also https://github.com/jipolanco/PencilFFTs.jl/issues/74


pxl-th:
Thanks!

We might also need to add other aliasing things & tests for it.
But that can be done later.

PR #679, opened 2024-09-30
ROCmDiscovery: Ensure all globals are defined

jpsamaroo:
This fixes a load-time error I get due to not having ROCm libraries available:
```
ERROR: InitError: UndefVarError: `libdevice_libs` not defined in `AMDGPU`
Suggestion: this global was defined as `AMDGPU.ROCmDiscovery.libdevice_libs` but not assigned a value.
Stacktrace:
  [1] functional(component::Symbol)
    @ AMDGPU ~/.julia/packages/AMDGPU/jLWP2/src/utils.jl:80
  [2] __init__()
    @ AMDGPU ~/.julia/packages/AMDGPU/jLWP2/src/AMDGPU.jl:210
...
```

Specifically, this doesn't remove the `@error` message that I get before this, but it just makes sure that `using AMDGPU` will still work without error even if the ROCm libraries aren't available.

### DIFF
--- a/src/discovery/discovery.jl
+++ b/src/discovery/discovery.jl
@@ -34,6 +34,15 @@ function get_device_libs(from_artifact::Bool; rocm_path::String)
     end
 end
 
+macro _ensure_defined(var)
+    mod = @__MODULE__
+    esc(quote
+        if !isdefined($mod, $(QuoteNode(Symbol(var))))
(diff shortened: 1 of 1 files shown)


pxl-th:
I think I got it fixed at some point. I'll close it.

Issue #683, opened 2024-10-05
GC less effective in AMDGPU than CUDA


evelyne-ringoot:
Creating a multitude of small copies for benchmarking slows AMDGPU.jl down a lot, something not observed in CUDA.jl. The solution for this specific code is to avoid allocations all together, but this is (maybe?) not possible with every type of code. (I also remember having had some issues with benchmarktools, but cannot manage to reproduce them right now) Sharing the code here for future reference:

```
using AMDGPU, BSON
n_values=(2 .^(1:14))
timings=zeros(2,length(n_values))

function mybelapsed(A, B)
   AMDGPU.rocBLAS.gemm('N','N',copy(A),copy(B))
   t=0.0
   k=0
   while (k<1e5 && t<1)
       Acpy=copy(A)
       Bcpy=copy(B)
       AMDGPU.synchronize()
       t+= @elapsed (AMDGPU.@sync AMDGPU.rocBLAS.gemm('N','N',Acpy,Bcpy))
       AMDGPU.synchronize()
       k+=1
    end
    return t/k
end

function mybelapsed2(A, B)
   AMDGPU.rocBLAS.gemm('N','N',copy(A),copy(B))
   t=0.0
   k=0
   Acpy=copy(A)
   Bcpy=copy(B)
   if(k<1e5 && t<1)
       AMDGPU.synchronize()
       t+= @elapsed (AMDGPU.@sync AMDGPU.rocBLAS.gemm('N','N',Acpy,Bcpy);)
       AMDGPU.synchronize()
       Acpy.=A
       Bcpy.=B
       k+=1
    end
    return t/k
end


for (i,n) in enumerate(n_values)
   A=ROCArray(rand(Float32,n,n));
   B=ROCArray(rand(Float32,n,n));
   println(n)
   timings[1,i]=mybelapsed(A,B)
   GC.gc()
   sleep(1)
   timings[2,i]=mybelapsed2(A,B)
   GC.gc()
   sleep(1)
   BSON.@save "AMD_matmul_bench.bson" timings
end
```

[...]

PR #684, opened 2024-10-05
Warning about benchmarktools

evelyne-ringoot:
Related to https://github.com/JuliaGPU/AMDGPU.jl/issues/683, proposing to add a warning regarding the use of benchmarktools

### DIFF
--- a/docs/src/profiling.md
+++ b/docs/src/profiling.md
@@ -71,3 +71,28 @@ wall duration is lower.
 
 Use `HIP_LAUNCH_BLOCKING=1` to synchronize immediately after launching GPU kernels.
 This will allow to pinpoint exact kernel that caused the exception.
+
+## Benchmarking
+
+Do not use benchmarktools.jl with AMDGPU as it creates copies, which gives inaccurate results. Instead use a custom allocation-avoiding benchmark function, e.g.:
+
+```julia
+    ...
+function mybelapsed2(f, A,B )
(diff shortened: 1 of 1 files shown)


luraess:
Thanks for the addition. It is indeed valuable to have some notes about potential limitations using e.g. BenchmarkTools with AMDGPU. One could however relax slightly the formulation as BenchmarkTools may slow down code in some cases but this may not be generalisable to all cases. Also, it would be valuable to have a self contained example and potentially a direct comparison good vs bad perf such that a user could directly run the suggested code snippet. This could and should be further adapted depending on the outcome of investigating #683 .

vchuravy:
> In addition, the same performance issues are observed when using benchmarktools, which relies on copies.

I am surprised by this statement. BenchmarkTools does not introduce copies

evelyne-ringoot:
> > In addition, the same performance issues are observed when using benchmarktools, which relies on copies.
> 
> I am surprised by this statement. BenchmarkTools does not introduce copies

My bad, I think I might've put something on benchmarktools that is not actually its fault. I thought I saw weird performance results with benchmarktoold before, but I cannot manage to reproduce them right now so closing this for now

luraess:
In some case, in the "past", I also had weird or failing AMDGPU test when using BenchmarkTools which forced me to things in a more manual fashion. I did no longer bump into such failing configs recently though...

PR #695, opened 2024-11-12
Use rocprof_v2 to implement Profiler for AMDGPU in Julia

vchuravy:
- **Prefer librocalution_hip over librocalution**
- **WIP: Profiler in Julia**


vchuravy:
Grrr of course there is now rocprofv3 which does look like an improvement https://rocm.docs.amd.com/projects/rocprofiler-sdk/en/latest/index.html 

but now the generation fails for me at least.

PR #706, opened 2024-12-03
[rocSOLVER] Update gesvd!

amontoison:


amontoison:
The failing tests are related to Julia 1.11.2.
We have the same issue in CUDA.jl.

PR #714, opened 2025-01-01
Add hint about ROCM_PATH to documentation

dermesser:
It took me quite some searching and experimenting to figure out how the Fedora rocm packages can be leveraged from Julia. I found the answer and would like to share it in the documentation - there seems to be a quite appropriate place for it (the one I'd been staring at before).


pxl-th:
Thanks!

PR #728, opened 2025-01-24
[rocFFT] Don't perform a copy for each call of FFT

amontoison:
Similar to the PR in CUDA.jl:
https://github.com/JuliaGPU/CUDA.jl/pull/2578


pxl-th (approved):

amontoison:
@pxl-th I just wanted to check and add test that the input `ROCArray` used in fft are never modified.
It should be fine if rocFFT follows the same rules than cuFFT and FFTW.

PR #731, opened 2025-02-05
Bump AcceleratedKernels to version 0.3

anicusan:
Making mapreduce and accumulate respect the `init` value by having a separate `neutral=GPUArrays.neutral_element` is a breaking change, hence the version bump.


pxl-th:
I've already bumped AK version in #730. Thanks for fixing that!

anicusan:
Ah, I didn't see that - moving quick! Thanks :)

PR #744, opened 2025-03-25
Update Project.toml

miniskar:
Need version updates for LLD_jll and LLVM_jll to build with latest Julia and LLVM.


pxl-th:
If you can rebase it on master, CI should pass.

FYI, we still don't fully support LLVM 18, so adding 19 is more for the future

pxl-th:
Thanks!

PR #745, opened 2025-03-26
Restrict `GPUCompiler < 1.3`

pxl-th:
Broken after https://github.com/JuliaGPU/GPUCompiler.jl/pull/668


luraess:
Why do we bumped patch from 5 to 7 skipping 6?

pxl-th:
> Why do we bumped patch from 5 to 7 skipping 6?

That was by mistake :)

PR #746, opened 2025-03-27
[rocSOLVER] Interface geblttrf_npvt and geblttrs_npvt

amontoison:
cc @sshin23

### DIFF
--- a/src/solver/highlevel.jl
+++ b/src/solver/highlevel.jl
@@ -212,6 +212,61 @@ for (fname, elty) in (
     end
 end
 
+for (fname, elty) in (
+    (:rocsolver_sgeblttrf_npvt, :Float32),
+    (:rocsolver_dgeblttrf_npvt, :Float64),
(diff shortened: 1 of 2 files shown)


pxl-th:
Tests fail because of incompatibility with GPUCompiler 1.3. Are you on latests commit?

Otherwise there's already a fix, but it will be merged in a bit

amontoison:
@pxl-th 
Yes, I rebased before doing these commits.
I try to test my PR on the `MI300x` that we have here at Argonne but I have these errors:
```shell
'gfx942' is not a recognized processor for this target (ignoring processor)
warning: sramecc 'On' was requested for a processor that does not support it!
```

pxl-th:
MI300x requires LLVM 18, so you can try with Julia 1.12, but I haven't tested AMDGPu with 1.12 at all yet.

luraess:
Can this be merged together with #816 ?

amontoison:
@luraess Yes, we can merge both PRs.

PR #747, opened 2025-03-28
 Regenerate wrappers for ROCm v6.3.3

amontoison:
Regenerate the Julia wrappers for the following libraries:
- rocsparse
- rocfft
- rocblas
- rocsolver


pxl-th:
Is there some functionality missing in the old wrappers?

> I am wondering if we should not replace all @ccall by @gc_ccall in the same time.

Why?

amontoison:
Because we could call the GC during a `@ccall` with this modification but it will be the default behaviour with Julia 1.12.

amontoison:
@pxl-th I updated the PR, all tests should passed now.
Yes, AMD added many routines during the last year in rocBLAS, rocSPARSE and rocSOLVER.
The generic API for rocSPARSE was quite incomplete before this PR.

pxl-th:
Thanks!

PR #758, opened 2025-05-06
Improve docs

pxl-th:


luraess:
Also, (unsure if relevant here given I do not see it in the current make file), but in other repos I've added the `[source]` section to the docs Project.toml to avoid having to use the `push!(LOAD_PATH,"../src/")`  that was previously needed (see e.g. https://github.com/PTsolvers/Chmy.jl/blob/main/docs/Project.toml)

pxl-th:
I think its only supported on 1.11+, while I'm targeting 1.10 for docs + tests as it is a LTS version for now.

pxl-th:
Here's preview: https://amdgpu.juliagpu.org/previews/PR758/

pxl-th:
> > Here's preview: https://amdgpu.juliagpu.org/previews/PR758/
> 
> Looks fancy! Maybe a preview issue, but AMDGPU logo does not render on the docs landing page

It is now :)

PR #764, opened 2025-05-18
Export more types

amontoison:
I wanted to use `StridedROCVector` in a project and I remarked that only `StridedROCMatrix` was exported.

This PR exports the same symbols than `CUDA.jl`.

### DIFF
--- a/src/AMDGPU.jl
+++ b/src/AMDGPU.jl
@@ -30,7 +30,9 @@ const LAUNCH_BLOCKING::Ref{Bool} = Ref{Bool}(false)
 export @roc, roc, rocconvert
 export HIPDevice, has_rocm_gpu
 export ROCArray, ROCVector, ROCMatrix, ROCVecOrMat
-export DenseROCArray, DenseROCVector, DenseROCMatrix, StridedROCMatrix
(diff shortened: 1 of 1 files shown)


luraess:
Canceled CI as it was running for over a day now while we still get the ComplexF16 related failures in Julia 1.12 (which seem unrelated to this PR). Feel free to merge it tho. One thing I don't get os that we have a 30min timeout set and still CI was now running for over a day.

luraess:
Thanks! Do you need a tag release @amontoison ?

amontoison:
I will do another PR beforehand.
I would like to update `update_wrap` so that it doesn't use host memory by default, and that this behavior is autodetected, like in CUDA.jl.
I need to use `lock=false` for device memory, which is not the most user-friendly keyword argument for that:
[https://github.com/mit-shin-group/BlockStructuredSolvers/blob/main/src/gpu\_seq.jl#L33-L37](https://github.com/mit-shin-group/BlockStructuredSolvers/blob/main/src/gpu_seq.jl#L33-L37)

I think I have the rights on the repository to tag a release, but I'll ask you otherwise.
Thank you!

PR #771, opened 2025-05-25
Update indexing.jl

amontoison:
Backport the same fix than https://github.com/JuliaGPU/CUDA.jl/pull/2554.
I have an index 0 after a call to `findall` and I suspect that it is the same issue.

### DIFF
--- a/src/kernels/indexing.jl
+++ b/src/kernels/indexing.jl
@@ -9,7 +9,7 @@ function Base.findall(bools::AnyROCArray{Bool})
     I = keytype(bools)
     indices = cumsum(reshape(bools, prod(size(bools))))
 
-    n = @allowscalar indices[end]
+    n = isempty(indices) ? 0 : @allowscalar indices[end]
     ys = ROCArray{I}(undef, n)
 
     if n > 0


amontoison:
I think we need a synchronize, the reproducer is quite simple:
```julia
using AMDGPU

lcon = -Inf * ones(Float64, 524288)
lcon = ROCVector(lcon)

ucon = zeros(Float64, 524288)
ucon = ROCVector(ucon)

findall(lcon .!= ucon)
```
The size of the output should be `524288` but it can be of a different size if we call it multiple times:
```julia
524288-element ROCArray{Int64, 1, AMDGPU.Runtime.Mem.HIPBuffer}
524226-element ROCArray{Int64, 1, AMDGPU.Runtime.Mem.HIPBuffer}
524164-element ROCArray{Int64, 1, AMDGPU.Runtime.Mem.HIPBuffer}
523978-element ROCArray{Int64, 1, AMDGPU.Runtime.Mem.HIPBuffer}
524256-element ROCArray{Int64, 1, AMDGPU.Runtime.Mem.HIPBuffer}:
```

@pxl-th @luraess

amontoison:
I added `AMDGPU.@sync` and I have the same issue.
I also tested in a script and some issue.
The GPU that I use is MI250: https://www.jlse.anl.gov/amd-gpu-mi250
I will try another AMD GPU.

luraess:
Could it be we need a sync before the `unsafe_free` as else we may free indices before them being used by the inner kernel?

luraess:
The issue seems to be the `cumsum`. From your MWE:
```
using AMDGPU

lcon = -Inf * ones(Float64, 524288)
lcon = ROCVector(lcon)

ucon = zeros(Float64, 524288)
ucon = ROCVector(ucon)

bools = lcon .!= ucon

julia> cumsum(bools)
524288-element ROCArray{Int64, 1, AMDGPU.Runtime.Mem.HIPBuffer}:
      1
      ⋮
 524226

julia> cumsum(bools)
524288-element ROCArray{Int64, 1, AMDGPU.Runtime.Mem.HIPBuffer}:
      1
      ⋮
 524164

julia> cumsum(bools)
524288-element ROCArray{Int64, 1, AMDGPU.Runtime.Mem.HIPBuffer}:
      1
      ⋮
 524288
```

luraess:
So the issue seems to point to AcceleratedKernels https://github.com/JuliaGPU/AMDGPU.jl/blob/93b575f38f0f44e58c8df0d86076d2cb90b927b8/src/kernels/accumulate.jl#L7

PR #772, opened 2025-05-25
Include AcceleratedKernels-0.4

anicusan:
There are no GPU-related breaking changes (only a kwarg removed from the multithreaded CPU backend), so both 0.3 and 0.4 can be kept.


luraess:
Actually, let's maybe wait on https://github.com/JuliaGPU/AcceleratedKernels.jl/issues/41 before tagging

PR #773, opened 2025-05-28
Add code bloc to issue template

luraess:


luraess:
This PR improves the formatting of bug report and suggests the user to print their command output in code bloc formatted region to avoid markdown interpretation of unicode.

PR #778, opened 2025-06-14
RFC: automatic discovery on Fedora

simeonschaub:
(Fedora 41/42 ships LLVM 18 so requires Julia 1.12)


luraess:
> [...] it would be nice if everything worked without setting `ROCM_PATH` manually.

Agree

pxl-th:
Superseded by #788

PR #783, opened 2025-06-23
Add memory barrier semantics to `sync_workgroup`

vchuravy:
KA.__synchronize also has memory barrier semantics and should be
equivalent to OpenCL's `barrier(CLK_LOCAL_MEM_FENCE)`.

x-ref: https://github.com/JuliaGPU/OpenCL.jl/pull/300

This fixes the data-races I observed in https://github.com/JuliaGPU/AcceleratedKernels.jl/issues/41#issuecomment-2995688972

### DIFF
--- a/src/ROCKernels.jl
+++ b/src/ROCKernels.jl
@@ -171,7 +171,8 @@ end
 # - https://github.com/llvm/llvm-project/blob/714b2fdf3a385e5b9a95c435f56b1696ec3ec9e8/libclc/opencl/lib/amdgcn/mem_fence/fence.cl#L28
 @device_override @inline function KA.__synchronize()
     # XXX(vchuravy): Determine semantics, but agreement sofar is LOCAL_MEM barrier 
-    ccall("llvm.amdgcn.s.waitcnt", llvmcall, Cvoid, (Cint,), 0xff) # memory barrier
+    # We want CLK_GLOBAL_MEM_FENCE | CLK_LOCAL_MEM_FENCE
+    ccall("llvm.amdgcn.s.waitcnt", llvmcall, Cvoid, (Cint,), 0) # memory barrier
     AMDGPU.Device.sync_workgroup()
 end
 
--- a/src/ROCKernels.jl
+++ b/src/ROCKernels.jl
@@ -166,13 +166,7 @@ end
 
 # Other.
 
-# x-ref:
-# - https://github.com/llvm/llvm-project/blob/5c22793eadd8758d589eafd1cbbb2897ab8b3c8b/libclc/opencl/lib/amdgcn/synchronization/barrier.cl#L11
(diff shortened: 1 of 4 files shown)


src/ROCKernels.jl
```diff
@@ -166,7 +166,12 @@ end
 
 # Other.
 
+# x-ref:
+# - https://github.com/llvm/llvm-project/blob/5c22793eadd8758d589eafd1cbbb2897ab8b3c8b/libclc/opencl/lib/amdgcn/synchronization/barrier.cl#L11
+# - https://github.com/llvm/llvm-project/blob/714b2fdf3a385e5b9a95c435f56b1696ec3ec9e8/libclc/opencl/lib/amdgcn/mem_fence/fence.cl#L28
 @device_override @inline function KA.__synchronize()
+    # XXX(vchuravy): Determine semantics, but agreement sofar is LOCAL_MEM barrier 
+    ccall("llvm.amdgcn.s.waitcnt", llvmcall, Cvoid, (Cint,), 0xff) # memory barrier
```
vchuravy, line 174:
```suggestion
    # We want CLK_GLOBAL_MEM_FENCE | CLK_LOCAL_MEM_FENCE
    ccall("llvm.amdgcn.s.waitcnt", llvmcall, Cvoid, (Cint,), 0) # memory barrier
```

vchuravy: https://github.com/JuliaGPU/OpenCL.jl/pull/338

luraess:
Why don't we need
```
ccall("llvm.amdgcn.s.waitcnt", llvmcall, Cvoid, (Cint,), 0) # memory barrier
```
in https://github.com/JuliaGPU/AMDGPU.jl/blob/f8ce0525814de87218f96486a1d63128c81c1ba3/src/device/gcn/synchronization.jl#L6-L7
as well? Or actually,, have it there and then keep the KA function as is?

vchuravy:
> Why don't we need

In some sense since the docs of that function state:

> Waits until all wavefronts in a workgroup have reached this call.

Which is what it does, and it doesn't imply any memory barrier semantics, where as KerneAbstractions.@synchronize does imply memory barrier semantics. 

The memory barriers shoould probably be easier to access, but AMDGPU.jl may want to use `UnsafeAtomics.fence` and allow for more flexibility than what KA promises.

vchuravy:
After talking with @gbaraldi yesterday, we decided that it is likely safer to upgrade `sync_workgroup` to follow HIP.

vchuravy:
For the curios we emit:

```
      fence syncscope("workgroup") seq_cst
     call void @llvm.amdgcn.s.barrier()
      fence syncscope("workgroup") seq_cst
```

Which lowers to on MI210:

```
	s_waitcnt lgkmcnt(0)
	s_barrier
	s_waitcnt lgkmcnt(0)
```

PR #790, opened 2025-07-01
Re-enable exception reporting

pxl-th:

PR #800, opened 2025-07-14
Re-work exception reporting (take 2)

### LINKED ISSUE #270: Improve exception reporting
A few things need improving:
- Exception flags are currently per-module, which is wrong. We should integrate with the kernel state mechanism from GPUCompiler and make these per-kernel.
- It would be great if we could have more fine-grained exception reporting, specifically allowing us to report which workgroups and lanes are throwing.
- Maybe we can implement a basic debugger on-device which would let us introspect arguments, locals, and per-lane state?

pxl-th:
What this exception reporting mechanism covers:
- type of the exception;
- thread & block which caused the exception.

**Does not** report the stacktraces as in #790, but it is cheaper.

Closes #270.

```julia
julia> using AMDGPU

julia> function ker!(x)
           x[0] = 1
           return
       end
ker! (generic function with 1 method)

julia> x = ROCArray(zeros(Int, 1));

julia> @roc ker!(x);

julia> AMDGPU.synchronize()
ERROR: GPU Kernel Exception:
BoundsError: Out-of-bounds array access
workitemIdx: (x = 0x00000001, y = 0x00000001, z = 0x00000001)
workgroupIdx: (x = 0x00000001, y = 0x00000001, z = 0x00000001)
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] throw_if_exception(dev::HIPDevice)
   @ AMDGPU ~/.julia/dev/AMDGPU/src/exception_handler.jl:149
 [3] synchronize(stm::HIPStream; blocking::Bool, stop_hostcalls::Bool)
   @ AMDGPU ~/.julia/dev/AMDGPU/src/highlevel.jl:40
 [4] synchronize
   @ ~/.julia/dev/AMDGPU/src/highlevel.jl:36 [inlined]
 [5] synchronize()
   @ AMDGPU ~/.julia/dev/AMDGPU/src/highlevel.jl:36
 [6] top-level scope
   @ REPL[5]:1
```


pxl-th:
This implementation seems to cause invalid IR excessively and I'm not yet sure what's causing it:

```julia
Reason: unsupported dynamic function invocation (call to throw_inexacterror(f::Symbol, ::Type{T}, val) where T @ AMDGPU.Device ~/.cache/julia-buildkite-plugin/depots/687d1932-34cc-406b-8aac-ab7952bcde26/packages/AMDGPU/SLMta/src/device/quirks.jl:31)
```

pxl-th:
Benchmarking [GaussianSplatting.jl](https://github.com/JuliaNeuralGraphics/GaussianSplatting.jl):
- before:
```julia
Benchmark for `1000` steps:
 42.344268 seconds (14.81 M allocations: 15.178 GiB, 1.06% gc time, 1.53% compilation time)
```
- this PR:
```julia
Benchmark for `1000` steps:
 50.338692 seconds (14.53 M allocations: 15.160 GiB, 1.43% gc time, 1.38% compilation time)
```

So better than #790, but still some work left to do to improve things.

pxl-th:
With latest commit there's almost no performance hit:

```julia
43.340191 seconds (14.50 M allocations: 15.159 GiB, 0.83% gc time, 1.61% compilation time)
```

Issue #770, opened 2025-05-25
ROCclr segfault when running Julia with threads


leios:
Nothing works when using Julia in threaded mode:
```
[leios@noema Fable.jl]$ julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.11.3 (2025-01-21)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> using AMDGPU

julia> AMDGPU.zeros(10)
10-element ROCArray{Float32, 1, AMDGPU.Runtime.Mem.HIPBuffer}:
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0

julia> 
[leios@noema Fable.jl]$ julia -t 2
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.11.3 (2025-01-21)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> using AMDGPU

julia> AMDGPU.zeros(10)
10-element ROCArray{Float32, 1, AMDGPU.Runtime.Mem.HIPBuffer}:
julia: /usr/src/debug/hip-runtime/clr-rocm-6.2.4/rocclr/os/os_posix.cpp:321: static void amd::Os::currentStackInfo(unsigned char**, size_t*): Assertion `Os::currentStackPtr() >= *base - *size && Os::currentStackPtr() < *base && "just checking"' failed.

[7772] signal 6 (-6): Aborted
in expression starting at none:0
unknown function (ip: 0x7087c8970624)
gsignal at /usr/lib/libc.so.6 (unknown line)
abort at /usr/lib/libc.so.6 (unknown line)
unknown function (ip: 0x7087c88fe4ea)
unknown function (ip: 0x70874990c108)
unknown function (ip: 0x708749918dc7)
unknown function (ip: 0x708749706285)
[...]

Issue #767, opened 2025-05-20
AMDGPU.versioninfo() crashes when using Archlinux


vchuravy:
Both on Julia 1.10 and 1.11 

```
julia> using AMDGPU

julia> AMDGPU.versioninfo()
[ Info: AMDGPU versioninfo
julia: /usr/src/debug/hip-runtime/hip-runtime-clr/hipamd/src/hip_code_object.cpp:1152: hip::FatBinaryInfo** hip::StatCO::addFatBinary(const void*, bool): Assertion `err == hipSuccess' failed.

[462812] signal 6 (-6): Aborted
in expression starting at REPL[6]:1
unknown function (ip: 0x7a58175a574c)
gsignal at /usr/lib/libc.so.6 (unknown line)
abort at /usr/lib/libc.so.6 (unknown line)
unknown function (ip: 0x7a58175334e2)
unknown function (ip: 0x7a578404bc6f)
unknown function (ip: 0x7a5664f7a608)
unknown function (ip: 0x7a58177452f6)
unknown function (ip: 0x7a58177453cc)
_dl_catch_exception at /lib64/ld-linux-x86-64.so.2 (unknown line)
unknown function (ip: 0x7a581774c0c8)
_dl_catch_exception at /lib64/ld-linux-x86-64.so.2 (unknown line)
unknown function (ip: 0x7a581774c4dd)
unknown function (ip: 0x7a581759fb33)
_dl_catch_exception at /lib64/ld-linux-x86-64.so.2 (unknown line)
unknown function (ip: 0x7a5817742568)
unknown function (ip: 0x7a581759f622)
dlopen at /usr/lib/libc.so.6 (unknown line)
ijl_load_dynamic_library at /cache/build/tester-amdci5-12/julialang/julia-release-1-dot-11/src/dlload.c:365
jl_get_library_ at /cache/build/tester-amdci5-12/julialang/julia-release-1-dot-11/src/runtime_ccall.cpp:45 [inlined]
jl_get_library_ at /cache/build/tester-amdci5-12/julialang/julia-release-1-dot-11/src/runtime_ccall.cpp:29
ijl_lazy_load_and_lookup at /cache/build/tester-amdci5-12/julialang/julia-release-1-dot-11/src/runtime_ccall.cpp:73
macro expansion at /home/vchuravy/src/AMDGPU/src/utils.jl:125 [inlined]
[...]

Issue #696, opened 2024-11-12
Coredump on using AMDGPU


sakura-nyaa:
Hoping somebody who understands HIP/ROCM better than me can help me understand whats going on here.
Using the version you get when you use "add AMDGPU" I get a core dump instantly.
By going into src/discovery/discovery.jl and moving
```
        global libMIOpen_path = get_library(lib_prefix * "MIOpen"; rocm_path)
```
up to the top (it needs to come before libhsa gets loaded. one line below and the coredumps return.):
```
    try
        global libMIOpen_path = get_library(lib_prefix * "MIOpen"; rocm_path)
        global libhsaruntime = if Sys.islinux()
            get_library("libhsa-runtime64"; rocm_path, ext="so.1")
        else
            ""
        end

        # Linker.
        lld_path = get_ld_lld(rocm_path; from_artifact=false,
            artifact_library=:LLD_jll, artifact_field=:lld_path)
        lld_artifact = false
        if isempty(lld_path)
            lld_path = get_ld_lld(rocm_path; from_artifact=true,
                artifact_library=:LLD_jll, artifact_field=:lld_path)
            lld_artifact = true
        end
        global lld_path = lld_path
        global lld_artifact = lld_artifact

        # HIP.
        global libhip = get_library(
            Sys.islinux() ? "libamdhip64" : "amdhip64"; rocm_path)

        # Check if opaque pointers are enabled and turn off artifacts.
        llvm_args = get(ENV, "JULIA_LLVM_ARGS", "")
        enabled_opaque_pointers = occursin("-opaque-pointers", llvm_args)
        from_artifact = (
            # Detect HIP version, which will influence what device libraries to use.
            (isempty(libhip) || Base.thisminor(_hip_runtime_version()) > v"5.4")
            && !enabled_opaque_pointers)

        # If ROCm 5.5+ - use artifact device libraries.
        global libdevice_libs = get_device_libs(from_artifact; rocm_path)

        # HIP-based libraries.
        global librocblas = get_library(lib_prefix * "rocblas"; rocm_path)
        global librocsparse = get_library(lib_prefix * "rocsparse"; rocm_path)
        global librocsolver = get_library(lib_prefix * "rocsolver"; rocm_path)
        global librocrand = get_library(lib_prefix * "rocrand"; rocm_path)
        global librocfft = get_library(lib_prefix * "rocfft"; rocm_path)
    catch err
        @error """ROCm discovery failed!
[...]

PR #802, opened 2025-07-17
Add FAQ entry for AMDGPU on Archlinux

vchuravy:
Since quite a few folks are running into issues on Archlinux (#696, #770, #767), I thought it would be good to point out the alternative shown in https://github.com/JuliaGPU/AMDGPU.jl/issues/770#issuecomment-3003954390


vchuravy:
@leios I had success switching packages. (I also needed libxm2-legacy).

PR #813, opened 2025-07-29
update hip wrappers

simeonschaub:
These were generated fully automatically with the modified script in
gen. It seems like these wrappers never got updated for ROCm 6, so e.g.
the printing of the GCN arch name was incorrect. This now covers a much
larger subset of the HIP API. We probably don't need it but it shouldn't
hurt to have wrappers available.


luraess:
LGTM. @pxl-th can you check as well. Thks!

PR #816, opened 2025-08-18
[rocSPARSE] Support more conversions

amontoison:


amontoison:
@luraess Not really, I just backported a few recent modifications from `CUDA.jl`.
I plan to replace the legacy API of rocSPARSE by the generic one and revamp the tests at this moment.

PR #817, opened 2025-08-27
support cooperative groups

simeonschaub:
We should probably wrap some more intrinsics for this to be more useful,
a simple `sync_grid` example works already.


simeonschaub:
I think this can be merged as-is. Being able to sync across a grid already opens up some interesting usecases and more features can always be added later

luraess:
Thanks!

Issue #815, opened 2025-08-13
Support GPUToolbox 0.3.0


amontoison:
It will be great to support this version of GPUToolbox like CUDA.jl.

PR #818, opened 2025-08-29
Try to support GPUToolbox 0.3

kshyatt:


kshyatt:
should hopefully address https://github.com/JuliaGPU/AMDGPU.jl/issues/815

simeonschaub:
> What should be breaking?

What I meant was that according to semver, going from 0.2 -> 0.3 designates breaking changes. This doesn't contain any, so it should have been a 0.2.x release

> May this interferes somehow with the fact that AMDGPU because of ROCm is somehow bound to specific LLVM versions?

I don't see how this could interfere, GPUToolbox.jl supports a broader range of LLVM.jl versions than AMDGPU.jl, so that should be fine

PR #823, opened 2025-09-04
fix CI: add JLD2 dependency

simeonschaub:
Seems like AMDGPU.jl was forgotten :'( https://github.com/JuliaGPU/GPUArrays.jl/pull/597#issuecomment-2980645354


luraess:
Thanks. Was unsure whether adding JLD2 deps to test was the way to go. But given this was done for other backends packages (as per https://github.com/JuliaGPU/GPUArrays.jl/pull/597#issuecomment-2980645354), we may just do the same as well.

PR #828, opened 2025-09-21
Add Julia 1.11 Enzyme job to pipeline

wsmoses:
Added a new job for Julia 1.11 with Enzyme testing in the pipeline.

Note that I believe the current failure in CI should be fixed now Enzyme-side


luraess:
Thanks! LGTM and test failure are indeed unrelated.

PR #832, opened 2025-10-06
Even more BLAS tests and fixes

kshyatt:
A bunch of stuff in `gemm_batched` needed fixing/tests and a few other things are now tested as well.

### DIFF
--- a/test/rocarray/blas.jl
+++ b/test/rocarray/blas.jl
@@ -51,7 +51,6 @@ end
         @test testf(norm, rand(T, m))
         @test testf(BLAS.asum, rand(T, m))
         @test testf(axpy!, Ref(rand()), rand(T, m), rand(T, m))
-        @test testf(axpy!, Ref(rand()), rand(T, m), 1:m-1, rand(T, m), 1:m-1)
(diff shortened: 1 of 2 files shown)


kshyatt:
the failing 1.10 test passes locally for me...

luraess:
> the failing 1.10 test passes locally for me...

I did not yet figure out what leads sometimes to failure in 1.10 recently...

test/rocarray/blas.jl
```diff
@@ -452,6 +495,20 @@ end
                         (bt == 'T' ? transpose(B[:, :, i]) : B[:, :, i])
                     @test C[:, :, i] ≈ c
                 end
+                A = [rand(T, 4, 4) for ix in 1:batch_count]
```
luraess, line 498:
```suggestion
                A  = [rand(T, 4, 4) for ix in 1:batch_count]
```

luraess:
Seems to be some partial overlap with #826 (for `trsm` and `trmm`) - maybe worth checking as well?

kshyatt:
That PR doesn't touch the tests, so I think the two should be compatible for now

luraess:
Thanks!

PR #839, opened 2025-10-21
Update discovery.jl with minimal approach to fix Windows discovery process

jbiffl:
Changes `rel_libdir` for Windows to "bin", otherwise leaves it as "" and leaves `lib_prefix` untouched to preserve Linux functionality.

Note: `amdhip64_6.dll` is in a subdirectory on Windows, but `libamdhip64.so` is top level to the ROCm path in Linux. This difference may cause issues for `libhip` in the future if it is ever modified to be consistent across platforms.


luraess:
Thanks @jbiffl !

Issue #854, opened 2025-11-30
Scalar indexing with multiplication between a scalar and the adjoint of ROC sparse matrices


ytdHuang:
## Questionnaire

1. Does ROCm works for you outside of Julia, e.g. C/C++/Python?
No idea

2. Post output of `rocminfo`.
I don't have a AMD GPU, I get the error while testing our package on JuliaGPU buildkite: https://buildkite.com/julialang/quantumtoolbox-dot-jl/builds/2484/steps/canvas?jid=019ad3d5-11f4-4c43-b1e9-e62e1dae4b2a

3. Post output of `AMDGPU.versioninfo()` if possible.
```

┌───────────┬──────────────────┬───────────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
--
│ Available │ Name             │ Version   │ Path                                                                                                              ⋯
├───────────┼──────────────────┼───────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
│     +     │ LLD              │ -         │ /opt/rocm-6.4.4/lib/llvm/bin/ld.lld                                                                               ⋯
│     +     │ Device Libraries │ -         │ /root/.cache/julia-buildkite-plugin/depots/0191c5ee-e683-4665-a03b-3f2068748b59/artifacts/5ad5ecb46e3c334821f54c1 ⋯
│     +     │ HIP              │ 6.4.43484 │ /opt/rocm-6.4.4/lib/libamdhip64.so                                                                                ⋯
│     +     │ rocBLAS          │ 4.4.1     │ /opt/rocm-6.4.4/lib/librocblas.so                                                                                 ⋯
│     +     │ rocSOLVER        │ 3.28.2    │ /opt/rocm-6.4.4/lib/librocsolver.so                                                                               ⋯
│     +     │ rocSPARSE        │ 3.4.0     │ /opt/rocm-6.4.4/lib/librocsparse.so                                                                               ⋯
│     +     │ rocRAND          │ 2.10.5    │ /opt/rocm-6.4.4/lib/librocrand.so                                                                                 ⋯
│     +     │ rocFFT           │ 1.0.32    │ /opt/rocm-6.4.4/lib/librocfft.so                                                                                  ⋯
│     +     │ MIOpen           │ 3.4.0     │ /opt/rocm-6.4.4/lib/libMIOpen.so                                                                                  ⋯
[...]

PR #861, opened 2025-12-05
Update support for sparse adjoint / transpose

luraess:
Following up on #859 

With the current status, I was hoping this would fix issues with sparse transpose (and possible adjoint) as in #854 which currently dispatch wrong and fail with scalar indexing, but seems there is still a missing bit.


luraess:
Closing this in favour of #937

PR #862, opened 2025-12-05
Update profiling docs: rocprof v1, v3 and roctx

Technici4n:
Hi, this PR is the result of trying to profile some Julia code on LUMI (default ROCm version 6.0.something, there is a build of 6.2.4 available, planned upgrade to 6.3 soon), and eventually succeeding after multiple days. Please let me know if any information is wrong and I will happily correct it. It would also be helpful if others with more experience could run the various commands on other versions of ROCm and see if the situation is different there.

Here is a test script that I used:
```jl
using AMDGPU

function rangePush(message)
    @ccall "libroctx64".roctxRangePushA(message::Ptr{Cchar})::Cint
end

function rangePop()
    @ccall "libroctx64".roctxRangePop()::Cint
end

N = 10000
mat = ROCArray(randn(N, N))
vec = ROCArray(randn(N))

tot = 0.0
for i in 1:10
    rangePush("Iteration $i")
    tot += sum(mat .* vec)
    rangePop()
end
println(tot)
```

PS: Given the existence of https://github.com/JuliaGPU/AMDGPU.jl/pull/801, I suppose that rocprofv3 is not expected to be working yet?

### DIFF
--- a/docs/src/tutorials/profiling.md
+++ b/docs/src/tutorials/profiling.md
@@ -79,7 +79,7 @@ rocprofv3 --output-format pftrace --hsa-trace --marker-trace --kernel-trace --me
 This will produce a number of `prof_output.pftrace` files which can be visualized
 using [Perfetto UI](https://ui.perfetto.dev/).
 
-`rocprofv3` is now recommended by ADM, however on ROCm 6.2.4 nothing seems to be reported.
+`rocprofv3` is now recommended by AMD, however on ROCm 6.2.4 nothing seems to be reported.
 
 #### Visualization of the results
 Here is an example of visualizing the `profile.jl` script above in Perfetto.
--- a/docs/src/tutorials/profiling.md
+++ b/docs/src/tutorials/profiling.md
@@ -2,10 +2,11 @@
 
 ## rocprof
 
-[rocprofv2](https://github.com/ROCm/rocprofiler?tab=readme-ov-file#rocprofiler-v2)
-allows profiling both HSA & HIP API calls (rocprof being deprecated).
+[rocprof](https://github.com/ROCm/rocm-systems/tree/develop/projects/rocprofiler)
+allows profiling HSA & HIP API calls, kernel launches, and more...
+Multiple major versions are available: `rocprof`, `rocprofv2` and `rocprofv3`.
 
-Let's profile simple copying kernel saved in `profile.jl` file:
+Let's profile a simple copying kernel saved in a `profile.jl` file:
 ```julia
 using AMDGPU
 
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/profiling.md
```diff
@@ -38,13 +39,52 @@ main(2^24)
...
 ```bash
-ENABLE_JITPROFILING=1 rocprofv2 --plugin perfetto --hip-trace --hsa-trace --kernel-trace -o prof julia ./profile.jl
+rocprofv3 --output-format pftrace --hsa-trace --marker-trace --kernel-trace --memory-copy-trace  -- julia ./profile.jl
 ```
 
-This will produce `prof_output.pftrace` file which can be visualized
+This will produce a number of `prof_output.pftrace` files which can be visualized
 using [Perfetto UI](https://ui.perfetto.dev/).
 
+`rocprofv3` is now recommended by ADM, however on ROCm 6.2.4 nothing seems to be reported.
```
luraess:
```suggestion
`rocprofv3` is now recommended by AMD, however on ROCm 6.2.4 nothing seems to be reported.
```

docs/src/tutorials/profiling.md
```diff
@@ -2,10 +2,11 @@
 
 ## rocprof
 
-[rocprofv2](https://github.com/ROCm/rocprofiler?tab=readme-ov-file#rocprofiler-v2)
-allows profiling both HSA & HIP API calls (rocprof being deprecated).
+[rocprof](https://github.com/ROCm/rocm-systems/tree/develop/projects/rocprofiler)
```
luraess:
```suggestion
[rocprof](https://rocm.docs.amd.com/projects/rocprofiler/en/latest/index.html)
```

docs/src/tutorials/profiling.md
```diff
@@ -69,6 +109,30 @@ wall duration is lower.
...
+
+rangePush("Section name")
+# Launch some kernels, call some functions, etc...
+rangePop()
+```
+
+While [ROCTX.jl](https://github.com/JuliaGPU/ROCTX.jl) aims to offer a Julia wrapper around it,
+it does not seem to be working yet. PRs welcome!
+(Note: the `ccall`s above do _not_ require ROCTX.jl to be loaded!)
+
```
luraess: Thanks for this addition!

Technici4n:
So you would say close this one and send the RocTX bits separately? Sorry I am currently very busy. 😓

luraess:
Ideally yes, given that the rocprof part was addressed mostly in #895 (incl. some reasoning why we may want to strip older versions...) Would be fantastic if you could send the ROCTX bits separately

PR #885, opened 2026-02-23
Improving memory requirements on AMDGPU

neoblizz:
@luraess reviews would be appreciated!

- Sets `eager_gc` to `true` by default.
- The `maybe_collect()` function proactively triggers `GC.gc(false)` when GPU memory pressure exceeds ~75%, preventing excessive pool growth.
- Allocation statistics and pool tracking.
- Background pool cleanup every 60 seconds.
- `reclaim()` function, supports trimming the pool and returns the bytes cleaned.

## Testing Plan & Summary

Ran all tests locally on gfx942 (MI300X):

- 16,717 passed
- 13 failed
- 13 errors
- 30 broken
- 16,773 total (5m 42.5s)

Test | Failure | Reason
-- | -- | --
wmma_tests | ArgumentError: invalid base 10 digit ':' in "942:sramecc+:xnack-" | Pre-existing bug parsing the gfx942 architecture string (Fixed: https://github.com/JuliaGPU/AMDGPU.jl/pull/885/changes/93fe6c6783708cbb3c2d95be726bd40d94139df7)
hip_rocarray/solver | dA \ dB ≈ Af \ Bf -- numerical inaccuracy in linear solvers | Unrelated rocSOLVER numerical precision issue (produces wildly wrong results like 1e7 vs 0.05)
gpuarrays/broadcasting | Float16 broadcast comparison failure | Should also be fixed with: https://github.com/JuliaGPU/AMDGPU.jl/pull/885/changes/93fe6c6783708cbb3c2d95be726bd40d94139df7

**Disclaimer:** Parts of this PR are developed using claude-4.6-opus AI.

### DIFF
--- a/src/runtime/memory/hip.jl
+++ b/src/runtime/memory/hip.jl
@@ -110,7 +110,7 @@ end
 
 const __pin_lock = ReentrantLock()
 const __pinned_memory = Dict{Ptr{Cvoid}, Int}()   # ptr => bytesize
-const __pin_count    = Dict{Ptr{Cvoid}, Int}()     # ptr => refcount
+const __pin_count = Dict{Ptr{Cvoid}, Int}()       # ptr => refcount
 
 struct HostBuffer <: AbstractAMDBuffer
     device::HIPDevice


src/AMDGPU.jl
```diff
@@ -184,6 +184,10 @@ function __init__()
 
     if functional(:hip)
         HIP.devices()
+        if isinteractive()
```
gbaraldi: This doesn't make too much sense. Why only when doing interactive workloads?

neoblizz:
> Can you split the arch thing into a separate PR?

https://github.com/JuliaGPU/AMDGPU.jl/pull/886

src/runtime/memory/hip.jl
```diff
@@ -39,22 +39,23 @@ function HIPBuffer(bytesize; stream::HIP.HIPStream)
...
             ptr = ptr_ref[]
             ptr == C_NULL && throw(HIP.HIPError(HIP.hipErrorOutOfMemory))
             return ptr
         catch err
-            # TODO rethrow if not out of memory error
-            @debug "hipMallocAsync exception. Requested $(Base.format_bytes(bytesize))." exception=(err, catch_backtrace())
+            # Only retry for OOM. Any other error (e.g. hipErrorLaunchFailure
+            # from a prior kernel exception surfacing here) must propagate
+            # immediately so the caller sees the real cause.
+            err isa HIP.HIPError || rethrow(err)
```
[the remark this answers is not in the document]
neoblizz: No, I read this as "If err is not a HipError rethrow it" and only the HipError exceptions continue, so things that were silently error'ing out before the alloc are not caught as part of the alloc and wrongfully attributed.

src/runtime/memory/hip.jl
```diff
@@ -105,6 +106,12 @@ function transfer!(dst::HIPBuffer, src::HIPBuffer, bytesize::Int; stream::HIP.HI
     return
 end
 
+## Host memory pinning state
+
+const __pin_lock = ReentrantLock()
+const __pinned_memory = Dict{Ptr{Cvoid}, Int}()   # ptr => bytesize
+const __pin_count    = Dict{Ptr{Cvoid}, Int}()     # ptr => refcount
```
luraess:
```suggestion
const __pin_count = Dict{Ptr{Cvoid}, Int}()       # ptr => refcount
```

neoblizz:
> Is there a need for updating anything in the doc wrt this PR (e.g. in here https://amdgpu.juliagpu.org/dev/api/memory).
> 
> Also, are the new additions causing any API changes that would require a major release?
> 
> Besides the pool statistics benchmark results report (which looks great), I do not see the test from [neoblizz#1](https://github.com/neoblizz/AMDGPU.jl/pull/1) actually merged in this PR branch. Do I overlook something? Also, besides the % of free, do you have any reports on the timing with the new memory management approach versus previous?

I have the test in the other branch, I will create a separate PR for that.

I did some testing on a Quantum Error Correction (QEC) workload -- there's no significant difference in runtime. If there's workloads you'd like me to run, I can run to make sure this holds true. Let me share raw latency results with this version and the one in `master` for QEC.

PR #890, opened 2026-03-10
Initial wrap of hipTENSOR library

kshyatt:
Mostly copied from cuTENSOR. Tests are failing with weird segfaults I don't understand, I'll try to keep debugging, but if anyone else wants to drop in and help I'd be very appreciative as I am quite unfamiliar with HIP.

### DIFF
--- a/src/AMDGPU.jl
+++ b/src/AMDGPU.jl
@@ -4,6 +4,8 @@ using Adapt
 using BFloat16s
 using CEnum
 using GPUCompiler
+using GPUToolbox
+using GPUToolbox: @enum_without_prefix
 using GPUArrays
 using GPUArrays: allowscalar
 using Libdl
@@ -128,6 +130,7 @@ include("kernels/reverse.jl")
 include("blas/rocBLAS.jl")
 include("solver/rocSOLVER.jl")
 include("sparse/rocSPARSE.jl")
+include("tensor/hipTENSOR.jl")
 include("rand/rocRAND.jl")
 include("fft/rocFFT.jl")
 include("dnn/MIOpen.jl")
--- a/src/AMDGPU.jl
+++ b/src/AMDGPU.jl
@@ -4,7 +4,6 @@ using Adapt
 using BFloat16s
 using CEnum
 using GPUCompiler
-using GPUToolbox
 using GPUToolbox: @enum_without_prefix
 using GPUArrays
 using GPUArrays: allowscalar
--- a/src/AMDGPU.jl
+++ b/src/AMDGPU.jl
@@ -4,7 +4,6 @@ using Adapt
 using BFloat16s
 using CEnum
 using GPUCompiler
-using GPUToolbox: @enum_without_prefix
 using GPUArrays
 using GPUArrays: allowscalar
 using Libdl
--- a/src/tensor/hipTENSOR.jl
+++ b/src/tensor/hipTENSOR.jl
@@ -14,8 +14,7 @@ using Printf: @printf
 
 export has_hiptensor
 
-const _initialized = Ref{Bool}(false)
-has_hiptensor() = _initialized[]
+has_hiptensor() = AMDGPU.functional(:hiptensor)
 
 # core library
 include("libhiptensor.jl")
--- a/src/tensor/hipTENSOR.jl
+++ b/src/tensor/hipTENSOR.jl
@@ -1,0 +1,61 @@
+module hipTENSOR
+
+using AMDGPU
+using AMDGPU
+using AMDGPU: @gcsafe_ccall, @checked, @enum_without_prefix, @debug_ccall
+using AMDGPU: Mem
+import AMDGPU: libhiptensor, HandleCache, HIP, library_state
+import AMDGPU.Mem: alloc_or_retry!
+import .HIP: HIPContext, HIPStream, hipStream_t
+
+using CEnum: @cenum
+
+using Printf: @printf
+
+export has_hiptensor
+
+const _initialized = Ref{Bool}(false)
+has_hiptensor() = _initialized[]
+
+# core library
+include("libhiptensor.jl")
+
+# low-level wrappers
+include("error.jl")
+#include("utils.jl")
+include("types.jl")
+include("operations.jl")
+
+# high-level integrations
+include("interfaces.jl")
+
+
+## handles
+
+function create_handle()
+    AMDGPU.functional(:hiptensor) || error("hipTENSOR is not available")
+
+    handle_ref = Ref{hiptensorHandle_t}()
+    hiptensorCreate(handle_ref)
+    handle_ref[]
+end
+
+const IDLE_HANDLES = HandleCache{HIPContext,hiptensorHandle_t}()
+
+lib_state() = library_state(
+    :hipTENSOR, hiptensorHandle_t, IDLE_HANDLES,
+    create_handle, hiptensorDestroy, (handle, stream) -> 0 )
+
+handle() = lib_state().handle
+stream() = lib_state().stream
+
+function version()
+    #=ver = Ref{Cint}()
+    hiptensor_get_version(handle(), ver)
+    major = ver[] ÷ 100000
+    minor = (ver[] ÷ 100) % 1000
+    patch = ver[] % 100=#
+    return VersionNumber(2, 2, 0)
+end
+
+end
--- a/src/tensor/hipTENSOR.jl
+++ b/src/tensor/hipTENSOR.jl
@@ -1,9 +1,10 @@
 module hipTENSOR
 
 using AMDGPU
-using AMDGPU
 using AMDGPU: @gcsafe_ccall, @checked, @enum_without_prefix, @debug_ccall
 using AMDGPU: Mem
+
+using BFloat16s: BFloat16
 import AMDGPU: libhiptensor, HandleCache, HIP, library_state
 import AMDGPU.Mem: alloc_or_retry!
 import .HIP: HIPContext, HIPStream, hipStream_t
--- a/src/tensor/hipTENSOR.jl
+++ b/src/tensor/hipTENSOR.jl
@@ -51,7 +51,7 @@ stream() = lib_state().stream
 function version()
     ver = hiptensorGetVersion()
     major = ver ÷ 10000
-    minor = (ver ÷ 100) % 10000
+    minor = (ver ÷ 100) % 100
     patch = ver % 100
     return VersionNumber(join((major,minor,patch), "."))
 end
--- a/src/tensor/hipTENSOR.jl
+++ b/src/tensor/hipTENSOR.jl
@@ -1,7 +1,7 @@
 module hipTENSOR
 
 using AMDGPU
-using AMDGPU: @gcsafe_ccall, @checked, @enum_without_prefix, @debug_ccall
+using GPUToolbox: @gcsafe_ccall, @checked, @enum_without_prefix, @debug_ccall
 using AMDGPU: Mem
 
 using BFloat16s: BFloat16
--- a/src/tensor/hipTENSOR.jl
+++ b/src/tensor/hipTENSOR.jl
@@ -22,7 +22,6 @@ include("libhiptensor.jl")
 
 # low-level wrappers
 include("error.jl")
-#include("utils.jl")
 include("types.jl")
 include("operations.jl")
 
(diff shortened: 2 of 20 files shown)


luraess:
cscs-ci run

src/utils.jl
```diff
@@ -147,6 +149,8 @@ function functional(component::Symbol)
         return !isempty(librocrand)
     elseif component == :rocfft
         return !isempty(librocfft)
+    elseif component == :hiptensor
```
luraess, line 153: This is a presence-only check, but unlike other lines, hipTENSOR is CDNA only and exits the process on load on RDNA, so it needs an architecture gate (gfx908/90a/94x/950) or a subprocess probe.

```
$ julia -e 'using Libdl; Libdl.dlopen("/opt/rocm/lib/libhiptensor.so")'
Cannot proceed: unsupported host device detected. Exiting.
exit code: 1
```

(gfx1100, ROCm 7.2.3. find_rocm_library only scans filenames, so using AMDGPU itself stays safe.)

test/runtests.jl
```diff
@@ -47,6 +47,7 @@ for (lib, pred) in [
     :rocfft    => name -> name == "hip_rocarray/fft",
     :rocrand   => name -> name == "hip_rocarray/random",
     :rocsolver => name -> name == "hip_rocarray/solver",
+    :hiptensor => name -> startswith(name, "hiptensor/"),
```
luraess, line 50: Because that check is presence-only, the RDNA (Radeon GPUs from Buildkite CI) runners will pick this suite up and get `exit(1)` at the first `hiptensorCreate`, killing the whole run rather than failing a testset.

src/tensor/hipTENSOR.jl
```diff
@@ -0,0 +1,61 @@
...
+const IDLE_HANDLES = HandleCache{HIPContext,hiptensorHandle_t}()
+
+lib_state() = library_state(
+    :hipTENSOR, hiptensorHandle_t, IDLE_HANDLES,
+    create_handle, hiptensorDestroy, (handle, stream) -> 0 )
+
+handle() = lib_state().handle
+stream() = lib_state().stream
+
+function version()
```
luraess, line 51: `hiptensorGetVersion()` is already wrapped, so this can be a real call, though the header encodes the value as `major * 10000 + minor * 100 + patch` rather than the `÷ 100000 / % 1000` in the commented-out block.

Worth putting after the arch gate: `versioninfo()` calls this, and only survives on RDNA today because the function does not ccall.

src/tensor/operations.jl
```diff
@@ -0,0 +1,456 @@
+const ModeType = AbstractVector{<:Union{Char, Integer}}
+
+# remove the HIPTENSOR_ prefix from some common enums,
+# as they're namespaced to the cuTENSOR module anyway.
```
luraess: Copy-paste from cuTENSOR, should read "the hipTENSOR module".

src/tensor/types.jl
```diff
@@ -0,0 +1,310 @@
+## data types
+
+# note that, unlike cuTENSOR, hipTENSOR wants a *complex* compute descriptor for complex
```
luraess: Maybe state it about hipTENSOR directly rather than against cuTENSOR ("hipTENSOR wants a complex compute descriptor for complex operands: ..."), and likewise for the two other cuTENSOR comparisons in this file at lines 50 and 166.

src/tensor/types.jl
```diff
@@ -0,0 +1,310 @@
...
+# lookup fail with a bare KeyError
+function default_compute_type(table, operation, eltypes)
+    haskey(table, eltypes) || throw(ArgumentError(
+        "hipTENSOR $(version()) does not support a tensor $operation with element types " *
+        join(eltypes, ", ") * ". Supported combinations are: " *
+        join(map(k -> join(k, ", "), sort!(collect(keys(table)); by=string)), "; ") * "."))
+    return table[eltypes]
+end
+
+# unlike cuTENSOR, where the compute descriptor is an opaque pointer, hipTENSOR's is a
```
luraess: Same here: maybe just keep "hipTENSOR's compute descriptor is a plain enum, so a Julia type maps directly onto a hiptensorComputeDescriptor_t value".

src/tensor/types.jl
```diff
@@ -0,0 +1,310 @@
...
+        return ComplexF64
+    else
+        throw(ArgumentError("Data type equivalent for hiptensorDataType type $T does not exist!"))
+    end
+end
+
+
+## plan
+
+# unlike cuTENSOR, hipTENSOR 2.2 does not copy the objects a plan is built from: the plan
```
luraess: Keep this one, it is the only thing explaining why hipTensorPlan carries refs; maybe just reword the opener so it stands on its own.

src/tensor/types.jl
```diff
@@ -0,0 +1,310 @@
...
+
+mutable struct hipTensor{T, N}
+    data::ROCArray{T, N}
+    inds::Vector{Int32}
+
+    function hipTensor{T, N}(data::ROCArray{T,N}, inds::Vector) where {T<:Number, N}
+        if length(inds) != N
+            throw(ArgumentError("The number of indices must match the number of dimensions of the data."))
+        end
+        if !iszero(UInt(pointer(data)) % HIPTENSOR_ALIGNMENT)
```
luraess: `HIPTENSOR_ALIGNMENT` is `UInt32(1)` (line 236), so this branch is unreachable, and `copy(data)` would not guarantee alignment even if it ran.

src/tensor/hipTENSOR.jl
```diff
@@ -0,0 +1,61 @@
...
+export has_hiptensor
+
+has_hiptensor() = AMDGPU.functional(:hiptensor)
+
+# core library
+include("libhiptensor.jl")
+
+# low-level wrappers
+include("error.jl")
+#include("utils.jl")
```
luraess: `src/tensor/utils.jl` is added empty, so this line and the file could both go?

src/AMDGPU.jl
```diff
@@ -4,6 +4,8 @@ using Adapt
 using BFloat16s
 using CEnum
 using GPUCompiler
+using GPUToolbox
+using GPUToolbox: @enum_without_prefix
 using GPUArrays
 using GPUArrays: allowscalar
```
luraess, line 8: Line 8 on its own is enough; the blanket import also pulls `i8/i16/i32/u8/u16/u32, @memoize, SimpleVersion, LazyInitialized` into AMDGPU's namespace.

test/hiptensor/base.jl
```diff
@@ -0,0 +1,29 @@
...
+        N = 2
+        dmax = 2^div(18,N)
+        dims = rand(2:dmax, N)
+        p = randperm(N)
+        indsA = collect(('a':'z')[1:N])
+        dimsA = dims
+        A = rand(Float64, dimsA...)
+        dA = ROCArray(A)
+        p = randperm(N)
+        indsA = collect(('a':'z')[1:N])
```
luraess, line 15: l.12-18: `p` and `indsA` are assigned twice and neither `p` nor `dmax` is used in this testset (leftover from the CUDA.jl version?)

test/runtests.jl
kshyatt: We'll need a check also for `has_hiptensor` then?

luraess:
Thanks for pushing this through, the wrapper reads well and the notes left about hipTENSOR's real behaviour are useful.

- The arch gate seems the blocker. hipTENSOR is CDNA only (Instinct accelerators) and exits the process on load on RDNA (Radeon, the Buildkite CI GPUs), while `functional(:hiptensor)` is a presence-only check, so on our runners this kills the test run instead of skipping it. Something we have to gate around it. The `version()` stub is the other one we should fix before merging.

- I'd reword some of the comments. The 2.2 quirk notes are load-bearing: the mode-order limitation is why `reordered_descriptor` exists, the swapped `opAB`/`opABC` explains a deliberate argument swap, and the "the plan does not copy what it is built from" note is why `hipTensorPlan` carries `refs`. Four of them are phrased against cuTENSOR, which is not needed and they could stand on their own.

Maybe the cuTENSOR deltas and the version-specific quirks should go in some succinct doc. There are no docs in this PR and `docs/src/libraries/` already has `dnn/fft/linalg/rand/sparse`, so a `tensor.md` with a "Known hipTENSOR 2.2 limitations" section would fit nicely, with short pointers at the call sites (also, easier to re-verify when ROCm bumps).

Also, `:hiptensor` is added to the functional docstring and dispatch but not to the `:all` tuple. Right or oversight?

luraess (changes requested):

test/runtests.jl
luraess: I guess so, yeah. We need to check if hipTensor is there, and if so if we are on CDNA arch.

kshyatt:
@luraess did the updates satisfy the concern you had?

src/tensor/hipTENSOR.jl
```diff
@@ -0,0 +1,59 @@
...
+    :hipTENSOR, hiptensorHandle_t, IDLE_HANDLES,
+    create_handle, hiptensorDestroy, (handle, stream) -> 0 )
+
+handle() = lib_state().handle
+stream() = lib_state().stream
+
+function version()
+    ver = hiptensorGetVersion()
+    major = ver ÷ 10000
+    minor = (ver ÷ 100) % 10000
```
luraess:
```suggestion
    minor = (ver ÷ 100) % 100
```
should be % 100 since the header defines the value as `major * 10000 + minor * 100 + patch`, minor is bounded at 99 and the mask has to be % 100.

src/utils.jl
```diff
@@ -162,6 +178,30 @@ function functional(component::Symbol)
     end
 end
 
+# hipTENSOR only ships kernels for the architectures listed in
+# `hiptensorSupportedArchitectures.cmake` of the ROCm install (Composable Kernel
```
luraess, line 182: ROCm ships this exact list as `hiptensorSupportedArchitectures.cmake` in the hipTENSOR cmake dir, so parsing it when present (falling back to this tuple for runtime-only installs) would track the install rather than drift each release. Could be worth doing because gfx940/gfx941 make the constant a superset of what 7.x actually supports, and superset may be unsafe as those devices would pass the gate and we would be back to dlopen deciding the outcome.

src/AMDGPU.jl
```diff
@@ -4,6 +4,7 @@ using Adapt
 using BFloat16s
 using CEnum
 using GPUCompiler
+using GPUToolbox: @enum_without_prefix
```
luraess: This using seems now dead. The hipTENSOR module imports the macro directly from GPUToolbox, and nothing in AMDGPU seems using it.

src/utils.jl
```diff
@@ -162,6 +178,30 @@ function functional(component::Symbol)
...
+(e.g. `"gfx90a"`; trailing target features as in `"gfx90a:sramecc+:xnack-"` are
+ignored) or for the architecture of `device`.
+
+This only inspects the architecture; use `AMDGPU.functional(:hiptensor)` to also
+check that the hipTENSOR library itself was found.
+"""
+hiptensor_supported(arch::AbstractString) =
+    first(split(arch, ':')) in HIPTENSOR_ARCHS
+
+hiptensor_supported(device) = hiptensor_supported(HIP.gcn_arch(device))
```
luraess:
```suggestion
hiptensor_supported(device::HIP.HIPDevice) = hiptensor_supported(HIP.gcn_arch(device))

```

luraess (changes requested):
Thanks. I left a few minor things but looks good. The one remaining is whether you'd be willing to add something to the docs, like `docs/src/libraries/tensor.md`. The in file comments on porting notes and 2.2 quirks are still spread across three files and having these listed into the docs could be helpful, maybe.

kshyatt:
Added some doc notes :)

luraess:
why is CI crying red... Seems missing a HIP import on MI300 and idk what BK is up to...

kshyatt:
Let me rebase on top of `main` and see if it was intermittent?

src/utils.jl
```diff
@@ -162,6 +178,31 @@ function functional(component::Symbol)
...
+(e.g. `"gfx90a"`; trailing target features as in `"gfx90a:sramecc+:xnack-"` are
+ignored) or for the architecture of `device`.
+
+This only inspects the architecture; use `AMDGPU.functional(:hiptensor)` to also
+check that the hipTENSOR library itself was found.
+"""
+hiptensor_supported(arch::AbstractString) =
+    first(split(arch, ':')) in HIPTENSOR_ARCHS
+
+hiptensor_supported(device::HIP.HIPDevice) = hiptensor_supported(HIP.gcn_arch(device))
```
luraess:
```suggestion
hiptensor_supported(device) = hiptensor_supported(HIP.gcn_arch(device))
```

luraess: @kshyatt reason why CI fails. I'd revert this suggestion. We loose the clean method Error but else we'd need to have this duplicated somewhere after HIP loads which is maybe not worth it.

luraess (approved):
LGTM now.

luraess:
Thanks for the addition!

PR #898, opened 2026-04-10
Store ROCArray offset in bytes instead of elements

maleadt:
The element-based offset was lossy when materializing reinterpret on views with non-aligned offsets (e.g., reinterpreting a view of Int32 as Int64). The byte offset would get truncated by integer division when converting to the new element count.

x-ref https://github.com/JuliaGPU/CUDA.jl/issues/2980


maleadt:
Merging this so that we can move forwards on the GPUArrays.jl version of this.

luraess:
Thanks!

PR #899, opened 2026-04-17
Test 1.13 and nightly

kshyatt:
Let these soft fail, but it's helpful to see the errors for upcoming releases


luraess:
LGTM. Failure on 1.10 and Enzyme relate to running on the CI runner with ROCm 7+ which has issues.

PR #905, opened 2026-05-08
Fix sparse matrix adaption

Abdelrahman912:
Right now, adaption of sparse matrix gives the same CPU matrix:

```
using AMDGPU, SparseArrays
import KernelAbstractions as KA, Adapt
A = spdiagm(0 => ones(5), -1 => -ones(4), 1 => -ones(4))
typeof(Adapt.adapt(ROCBackend(), A))   # print SparseMatrixCSC, not ROCSparseMatrixCSC
```

### DIFF
--- a/src/ROCKernels.jl
+++ b/src/ROCKernels.jl
@@ -23,7 +23,7 @@ function KA.device!(kab::ROCBackend, id::Int)
     return
 end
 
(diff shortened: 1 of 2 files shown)


luraess:
Thanks for the fix.

PR #906, opened 2026-05-09
Add `SparseMatricesCSR.jl` Ext

Abdelrahman912:
Add support to CPU CSR matrices, e.g., `ROCSparseMatrixCSR(::SparseMatrixCSR)`


luraess:
Thanks for the contribution! The structure closely mirrors what CUDA.jl does in `lib/cusparse/ext/SparseMatricesCSRExt.jl`.

One thing to consider (unless I am wrong): `SparseMatricesCSR.SparseMatrixCSR{Bi,Tv,Ti}` has a type parameter `Bi` for the index base (0 or 1). The CPU→GPU constructors currently use `Mat.rowptr`/`Mat.colval` directly without checking `Bi`, which would silently produce incorrect index arrays for a 0-based `SparseMatrixCSR{0}`; ROCSPARSE expects 1-based pointers in the Julia wrappers (matching the `SparseMatrixCSR{1}` used in the GPU→CPU direction). Restricting the dispatch to `{1}` would turn this into a clear error instead:

```julia
ROCSparseMatrixCSR{T}(Mat::SparseMatrixCSR{1}) where {T} = ...
```

_Note that CUDA.jl has the same gap, but this could be also fixed there if needbe._

Abdelrahman912:
This should be ready for merge.

luraess:
Thanks for the update. There are failing tests now. Upon fixing those we could then merge, if you can have a look.

test/hip_rocsparse/sparse_matrices_csr.jl
```diff
@@ -0,0 +1,31 @@
...
+                end
+                for ROCSparseMatrixType2 in (ROCSparseMatrixCSC, ROCSparseMatrixCSR, ROCSparseMatrixCOO, ROCSparseMatrixBSR)
+                    ROCSparseMatrixType1 == ROCSparseMatrixType2 && continue
+                    dA2 = ROCSparseMatrixType2 == ROCSparseMatrixBSR ? ROCSparseMatrixType2(dA1, blockdim) : ROCSparseMatrixType2(dA1)
+                    @testset "conversion $ROCSparseMatrixType1 --> $ROCSparseMatrixType2" begin
+                        @test collect(dA1) ≈ collect(dA2)
+                    end
+                end
+            end
+        end
```
Abdelrahman912, line 30:
```
Error During Test at /var/lib/buildkite-agent/builds/amdgpu1-luraess-com/julialang/amdgpu-dot-jl/test/hip_rocsparse/sparse_matrices_csr.jl:12
  Got exception outside of a @test
  MethodError: no method matching (AMDGPU.rocSPARSE.ROCSparseMatrixBSR{Float64})(::AMDGPU.rocSPARSE.ROCSparseMatrixCSC{Float64, Int32}, ::Int64)

  Closest candidates are:
    (AMDGPU.rocSPARSE.ROCSparseMatrixBSR{Float64})(::AMDGPU.rocSPARSE.ROCSparseMatrixCSR{Float64}, ::Integer; dir, inda, indc)
     @ AMDGPU /var/lib/buildkite-agent/builds/amdgpu1-luraess-com/julialang/amdgpu-dot-jl/src/sparse/conversions.jl:224
    (AMDGPU.rocSPARSE.ROCSparseMatrixBSR{T})(::SparseArrays.SparseMatrixCSC, ::Any) where T
     @ AMDGPU /var/lib/buildkite-agent/builds/amdgpu1-luraess-com/julialang/amdgpu-dot-jl/src/sparse/array.jl:410
    (AMDGPU.rocSPARSE.ROCSparseMatrixBSR{T})(::SparseMatricesCSR.SparseMatrixCSR{1}, ::Any) where T
     @ AMDGPUSparseMatricesCSRExt /var/lib/buildkite-agent/builds/amdgpu1-luraess-com/julialang/amdgpu-dot-jl/ext/AMDGPUSparseMatricesCSRExt.jl:26
    ...

  Stacktrace:
    [1] AMDGPU.rocSPARSE.ROCSparseMatrixBSR(x::AMDGPU.rocSPARSE.ROCSparseMatrixCSC{Float64, Int32}, blockdim::Int64)
      @ AMDGPU.rocSPARSE /var/lib/buildkite-agent/builds/amdgpu1-luraess-com/julialang/amdgpu-dot-jl/src/sparse/array.jl:417
```

Probably this is a side tangent to this PR, but it's just so happen that it's the main culprit of the CI fail, due to this test nested loop which essentially converts every sparse matrix to every other sparse matrix. That said,
@luraess is there a reason why there is no conversion between `CSC` & `BSR` ? Like this one: 
https://github.com/JuliaGPU/CUDA.jl/blob/54c758682d909f32d656fcbe8c43c20062850927/lib/cusparse/src/conversions.jl#L710-L711

If no need for them, I can just update the test, otherwise I can add those conversions, which hopefully will make CI happy.

luraess: > otherwise I can add those conversions

That would be the best way if you're willing to do so. Thanks!

luraess:
Thanks!

PR #907, opened 2026-05-09
Adapt to GPUCompiler changes

luraess:


luraess:
Further tweaks, x-ref: https://github.com/JuliaGPU/CUDA.jl/pull/3131

luraess:
Adding some context:
Seems the two pass are needed currently to (1) link libs before optimisation passes. Device lib functions (especially from `ocml`) need to be present here so the optimiser can inline them. (2) the finish module runs after optimisation/deferred codegen where new symbols are introduced between the two stages by the GPUCompiler runtime. Those weren't in the module during the first pass, so they need a second linking pass. The first pass feeds functions into the optimizer. The second pass catches symbols the runtime introduced after the optimizer ran. Diffs from CUDA.jl: maybe CUDA's runtime doesn't introduce new references to libdevice symbols after `link_libraries!` runs? Which allows the single pass + lazy linking approach?

luraess:
Merging this for now as to fix the warnings and CI failures. Further tweaks could follow up #917 .

PR #911, opened 2026-05-12
Skip test for missing HIP libs

luraess:
AMDGPU detects and warns about missing libs, but tests ignored those before and would fail.


luraess:
CI failures are not related

PR #915, opened 2026-05-19
Rework stream finalizer

luraess:
With HIP 7, passing an invalid `hipStream_t` to stream-related APIs causes a segfault rather than returning `hipErrorContextIsDestroyed`. This adds guards for it in both the finalizer and `isdone()` with an `isvalid` check, and ensure `valid` is set to `false` atomically _before_ `hipStreamDestroy` is called (so any concurrent reader sees the stream as invalid before the handle is freed).


luraess:
CI failure relates to #914 and are not related to this PR.

PR #918, opened 2026-05-20
Add performance tests

luraess:
Implementing performance regression tests in AMDGPU that are already present in other GPU backends.


luraess:
Benchmark runs succeed and logs suggest benchmark get uploaded https://buildkite.com/julialang/amdgpu-dot-jl/builds/3730#019e4e96-d7c2-45ad-ac52-a58ae794cd03/L442

However, seems nothing ends up (yet?) in https://amdgpu.juliagpu.org/bench/

luraess:
Merging this to trigger the gh-pages display and allow for further tweaks.

PR #919, opened 2026-05-21
RFC: automatic therock artifact selection

simeonschaub:
This is a proof of concept for automatically downloading ROCm libraries via the artifact system. Not sure this is the best approach, maybe we should wrap these tarballs as JLLs through Yggdrasil instead?

It's currently partially vibe coded and Linux only, since I couldn't find a reliable way of querying the gfx_target_version on Windows.


gbaraldi:
Have you seen that multiarch tarballs are available. Though I think they are gigantic

simeonschaub:
Yes, I saw them, but like you said I don't think it's great to ship such huge tarballs to users

simeonschaub:
This is now using the clang shipped by rocm as also proposed by @vchuravy in https://github.com/JuliaGPU/AMDGPU.jl/issues/931#issuecomment-4752687409. Locally I am getting miscompilations in the triangular matmul tests, which I will try to reduce and open an upstream issue.

gbaraldi:
Only thing is I wonder if you can split therock from using clang just to reduce the diff

simeonschaub:
Yes, I agree that makes sense. We'll have to figure out how this should interact with https://github.com/JuliaGPU/GPUCompiler.jl/pull/857 as well, since with those changes we'd have 3 possible backends - in tree, AMDGPU_LLVM_Backend_jll, and clang from the rocm path. What I like about the latter approach is that the compiler, ld.lld, device libs, and the other vendor libs always match and we don't have to worry about keeping all those different jlls up to date and in sync.

simeonschaub:
Summarizing our discussion:

- We add device libs to AMDGPU_Backend_jll
- We downgrade those device libs on the user side
- AMDGPU_Backend_jll keeps using upstream LLVM, we just fetch the device libs from ROCm's llvm, but compile them using Clang_jll as a HostBuildDependency
- We repackage therock artifacts to only contain the vendor libs like rocBLAS, etc., there's a lot in there we don't actually need
- We should add a ROCm version in the augmentation step, so in the future we can keep supporting devices dropped by therock

amontoison:
@simeonschaub It was on my todo list for 2027. We should maybe do a meeting to discuss about it.
I can request a custom packaging for Julia in TheRock.

simeonschaub:
> It was on my todo list for 2027. We should maybe do a meeting to discuss about it.

Sure, would love to! @gbaraldi, @luraess, @maleadt and I discussed this extensively during JuliaCon and came up with this plan

> I can request a custom packaging for Julia in TheRock.

So as it turns out, MIOpen actually requires `libllvm`, `libclang`, and I believe even the device libs to be present, so in #1024 I actually moved to again shipping the full therock tarballs instead of repackaging in Yggdrasil, since we need pretty much all of it

PR #923, opened 2026-05-26
Fix hardware FP atomics on Julia 1.13 (LLVM 20)

luraess:
In LLVM 20, `shouldExpandAtomicRMWInIR` for AMDGPU gained a legality check (`globalMemoryFPAtomicIsLegal`) that gates whether FP `atomicrmw` instructions are lowered to native hardware instructions or expanded to CAS loops. For targets like gfx1100 (RDNA3) that lack `AgentScopeFineGrainedRemoteMemoryAtomics`, the check requires `!amdgpu.no.fine.grained.memory` per-instruction metadata to return `true`. Without it, LLVM expands every FP atomic to a `global_atomic_cmpswap_b32` CAS loop regardless of the `amdgpu-unsafe-fp-atomics` function attribute.

The fix adds an empty `!amdgpu.no.fine.grained.memory` metadata node to all FP `atomicrmw` instructions (`fadd`/`fsub`/`fmax`/`fmin`) in `finish_module!` when `unsafe_fp_atomics` is enabled. This runs before the `AtomicExpandPass`, so LLVM 20 selects native instructions (e.g. `global_atomic_add_f32`) as expected. Adding the metadata is a no-op on CDNA targets and on LLVM 18 (Julia 1.12), which don't consult it.

### DIFF
--- a/src/compiler/codegen.jl
+++ b/src/compiler/codegen.jl
@@ -91,6 +91,19 @@ function GPUCompiler.finish_module!(
         end
     end
 
+    # LLVM 20+ requires !amdgpu.no.fine.grained.memory per-instruction metadata
+    # on FP atomicrmw to select native hardware atomics (e.g. global_atomic_add_f32)
+    # instead of expanding to a CAS loop on some targets (e.g. gfx1100).
+    if job.config.params.unsafe_fp_atomics
+        fp_binops = (LLVM.API.LLVMAtomicRMWBinOpFAdd, LLVM.API.LLVMAtomicRMWBinOpFSub,
+                     LLVM.API.LLVMAtomicRMWBinOpFMax, LLVM.API.LLVMAtomicRMWBinOpFMin)
+        empty_md = MDNode(Metadata[])
+        for fn in LLVM.functions(mod), bb in LLVM.blocks(fn), inst in LLVM.instructions(bb)
+            inst isa LLVM.AtomicRMWInst && LLVM.binop(inst) ∈ fp_binops &&
+                (LLVM.metadata(inst)["amdgpu.no.fine.grained.memory"] = empty_md)
+        end
+    end
+
     return entry
 end
 
--- a/src/compiler/codegen.jl
+++ b/src/compiler/codegen.jl
@@ -91,16 +91,22 @@ function GPUCompiler.finish_module!(
         end
     end
 
-    # LLVM 20+ requires !amdgpu.no.fine.grained.memory per-instruction metadata
-    # on FP atomicrmw to select native hardware atomics (e.g. global_atomic_add_f32)
-    # instead of expanding to a CAS loop on some targets (e.g. gfx1100).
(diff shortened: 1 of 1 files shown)


luraess:
This is the best I could come up with the help from Claude. Any feedback welcome to see if one could address this in a better fashion.

vchuravy:
How is Clang handling this? 

We are emitting them here https://github.com/JuliaConcurrent/UnsafeAtomics.jl/blob/master/ext/UnsafeAtomicsLLVM/atomics.jl I think, but there is currently no special handling for AMDGPU.

There is also `‘amdgpu.no.remote.memory’ `

I am a bit worried about universal emission

https://rocm.docs.amd.com/en/docs-6.2.0/conceptual/gpu-memory.html#coherence

fine-grained has a meaning in terms of coherence and I am not sure when/if we are getting fine-grained memory on all systems.

luraess:
Seems that Clang's `setTargetAtomicMetadata` in AMDGPU.cpp applies `!amdgpu.no.fine.grained.memory` unconditionally on all FP atomics whenever `allowAMDGPUUnsafeFPAtomics()` is true. 

The semantic contract is on the `unsafe_fp_atomics` opt-in: by enabling it, the caller asserts their memory is not fine-grained. Note that Clang would also add `!amdgpu.ignore.denormal.mode` for `fadd float` specifically.

- CGAtomic.cpp: https://github.com/llvm/llvm-project/blob/llvmorg-20.1.0/clang/lib/CodeGen/CGAtomic.cpp#L2078
- AMDGPU.cpp: https://github.com/llvm/llvm-project/blob/llvmorg-20.1.0/clang/lib/CodeGen/Targets/AMDGPU.cpp#L565

luraess:
## Summary

LLVM 20 changed how `AtomicExpandPass` decides whether to emit native FP atomic instructions (e.g. `global_atomic_add_f32`) or fall back to a compare-and-swap loop. For targets like **gfx1100** that lack `AgentScopeFineGrainedRemoteMemoryAtomics`, the backend now requires a per-instruction `!amdgpu.no.fine.grained.memory` metadata node on each `atomicrmw`. The function-level `amdgpu-unsafe-fp-atomics` attribute alone is no longer sufficient — it is marked `// TODO: Remove` in LLVM source.

### Fix

In `finish_module!`, when `unsafe_fp_atomics=true`, annotate all FP `atomicrmw` instructions (`fadd`, `fsub`, `fmax`, `fmin`) with:

- `!amdgpu.no.fine.grained.memory` — on all FP RMW ops
- `!amdgpu.ignore.denormal.mode` — additionally on `fadd float` only

This mirrors exactly what Clang does in `AMDGPUTargetCodeGenInfo::setTargetAtomicMetadata` ([source](https://github.com/llvm/llvm-project/blob/llvmorg-20.1.0/clang/lib/CodeGen/Targets/AMDGPU.cpp#L565)).

The `unsafe_fp_atomics` opt-in is the safety contract: the caller asserts the memory is device-local and not CPU-coherent host memory.

PR #927, opened 2026-06-02
Add device libs from ROCm 7.0.2 for Julia 1.13

luraess:
Allow to use device libs jll from ROCm 7.0.2 with Julia 1.13 (both using LLVM 20)


simeonschaub (approved):

luraess:
Still working on this. It turns out that there were issues with device libs 6.2.1 and another test failure.

PR #929, opened 2026-06-10
Support for WMMA instructions for RDNA4 GPUs

ffrancesco94:
As per title. Most things work in the same way as RDNA3, except for the fact that you don't need data duplication as RDNA4 lanes have 8 elements and not 16 (8*2 as it was before). I kept both implementation and tests separate from the RDNA3 version, but maybe in the future they could be kind of merged with a runtime dispatch based on the hardware.

DISCLAIMER: I did use Mistral vibe to get the first draft and then took development from there. I do not know if there is a specific policy against use of AI for pull requests to this project, thus I understand if you're not willing to look at it. The tests pass on my RX 9070XT and even if you diff with the RDNA3 version you'll see that the logic is the same and what changes is the shape of the fragments and addressing in the lanes.


ffrancesco94:
My last commits fail the CI but if I look at the logs there are no errors actually, I don't know if it's related to the updates to the buildkite.

luraess:
Yes, we had issues with the runners. I relaunched the failed jobs. Would the current status be ready for review or are you still working on it?

ffrancesco94:
If it passes the CI I'm happy to get it reviewed!

luraess:
The addition looks good to me. I was curious if you had specific idea in mind on how to achieve
> but maybe in the future they could be kind of merged with a runtime dispatch based on the hardware.

ffrancesco94:
I haven't given it much thought, but something along the lines of a trait, where we have empty structs RDNA4 and RDNA3 as subtypes of a `RDNAArch` abstract struct and we redefine mma to take also this struct as an argument so that it can dispatch to the right intrinsics. Then what we actually export is a `mma` method which checks the architecture and calls the appropriate version. I don't know if it makes sense, I haven't really tried because I only have a RDNA4 GPU so I can't test this dispatch myself (and the buildkite runner only seems to have a RDNA3 from what I can see) and also because it was my first contribution and I didn't want to mess up too much with the existing codebase. When I get more comfortable I'll try to get into that :)

luraess:
Thanks, feel free to open an issue reporting the above thoughts so we could have a trace for future todos.

One last thing, would it make sense to explicitly have `WMMA_RDNA3` (by analogy to now `WMMA_RDNA4` and possible future ones) instead of `WMMA` which stands for RDNA_3 only as to be consistent?

ffrancesco94:
It makes sense and I have changed that. I just hope it doesn't break anything downstream for the people who were referring to `WMMA` directly. Buildkite running now.

test/wmma_tests.jl
```diff
@@ -2,27 +2,28 @@ using Test
 using AMDGPU
 using AMDGPU: ROCArray, @roc
 using BFloat16s
-using AMDGPU.Device: WMMA, workitemIdx, workgroupIdx
+using AMDGPU.Device: WMMA_RDNA3_RDNA3, workitemIdx, workgroupIdx
```
luraess:
```suggestion
using AMDGPU.Device: WMMA_RDNA3, workitemIdx, workgroupIdx
```
I guess this is a typo?

luraess:
@pxl-th is it fine for you to rename the WMMA -> WMMA_3 for consistency with 4?

test/wmma_tests.jl
ffrancesco94: Yep, that's why the tests were failing

luraess:
I am planing in merging this soon but we will need to tag a major version bump given we're breaking the API renaming WMMA -> WMMA_3. So I'll do this right before release.

ffrancesco94:
Yes, completely understandable. Thank you for the review by the way. I hope to be able to contribute more and more in the upcoming months!

luraess:
Thanks!

PR #932, opened 2026-06-19
Add io stream to versioninfo

luraess:
This PR adds an io stream to `AMDGPU.versioninfo()` to be compatible with KA's interface and implements `KA.versioninfo`.

x-ref https://github.com/JuliaGPU/KernelAbstractions.jl/issues/617#issuecomment-4742139302 and https://github.com/JuliaGPU/KernelAbstractions.jl/pull/712

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -21,7 +21,7 @@ function versioninfo(io::IO=stdout)
 
     if functional(:hip)
         println(io)
-        @info "AMDGPU devices"
+        println(io, "AMDGPU devices")
         show(io, MIME"text/plain"(), AMDGPU.devices())
         println(io)
     end
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1,4 +1,4 @@
-function versioninfo()
+function versioninfo(io::IO=stdout)
     @info "AMDGPU versioninfo"
     _status(st::Bool) = st ? "+" : "-"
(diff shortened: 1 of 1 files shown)


luraess:
cscs-ci run

src/utils.jl
```diff
@@ -15,14 +15,15 @@ function versioninfo()
...
 
-    PrettyTables.pretty_table(data; column_labels=[
+    PrettyTables.pretty_table(io, data; column_labels=[
         "Available", "Name", "Version", "Path"],
         alignment=[:c, :l, :l, :l])
 
     if functional(:hip)
-        println()
+        println(io)
         @info "AMDGPU devices"
```
simeonschaub: `@info` will still always print to stdout, right? Can probably just use print instead

luraess: What are you suggesting?

simeonschaub: Something like:
```suggestion
        println(io, "AMDGPU devices")
```

src/utils.jl
```diff
@@ -1,4 +1,4 @@
-function versioninfo()
+function versioninfo(io::IO=stdout)
     @info "AMDGPU versioninfo"
```
simeonschaub:
```suggestion
    println(io, "AMDGPU versioninfo")
```

PR #934, opened 2026-06-19
Improve discovery robustness

luraess:
Ubuntu 26 ships ROCm to be installed with `apt` in a location AMDGPU does not yet expects and with modified LLD naming which causes issues https://github.com/JuliaGPU/AMDGPU.jl/issues/920#issuecomment-4536656216.

### DIFF
--- a/src/discovery/utils.jl
+++ b/src/discovery/utils.jl
@@ -9,6 +9,13 @@ function check_rocm_path(path::String)
     isfile(joinpath(path2, libname)) && @goto success
     path2 = joinpath(path, "lib64")
     isfile(joinpath(path2, libname)) && @goto success
+    # Check Debian/Ubuntu multiarch lib paths (e.g. /usr/lib/x86_64-linux-gnu)
+    if Sys.islinux()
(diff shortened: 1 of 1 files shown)


luraess:
Ubuntu 26.04 installs ROCm 7.1.0 into `/usr/lib/x86_64-linux-gnu/` (Debian multiarch) rather than `/opt/rocm/`, causing all ROCm components to be reported as unavailable.

Two fixes:

- **`check_rocm_path`**: probe Debian/Ubuntu multiarch subdirectories (`lib/<triplet>`) so `check_rocm_path("/usr")` resolves correctly on Ubuntu 26.04.
- **`find_ld_lld`**: replace split-by-position version parsing with a regex, fixing a warning caused by Ubuntu's lld emitting `"Ubuntu LLD 21.0.0 ..."` instead of `"AMD LLD X.Y.Z ..."`.

luraess:
Merging this as the improvements are independent of the issues in #920 and improve robustness anyway. We can further iterate on the issue nevertheless.

PR #935, opened 2026-06-19
Fix rocFFT error with multiple tasks

luraess:
### Problem

Running FFT operations on a `Threads.@spawn`-ed task (or any task other than the one that created the plan) would intermittently produce `rocfft_status_failure`. This was observed in CI when running tests in parallel on MI300.

The root cause is a constraint in the rocFFT API: `rocfft_execution_info_set_stream` must be called on the same OS thread as the subsequent `rocfft_execute`. Julia tasks can migrate between OS threads at yield points, so storing a single `rocfft_execution_info` per plan and updating its stream before execution is not sufficient — the task may land on a different OS thread between the stream update and the actual execute call. rocFFT's OS-thread constraint applies to the entire `create → set_stream → execute` sequence.

### Fix

`rocfft_execution_info` is now created fresh per execution inside each `unsafe_execute!` call, rather than being stored in the plan struct. The stream and work buffer are set immediately before `rocfft_execute`, with no yield points between setup and execution. The `execution_info` is destroyed after each call.

This matches the intended C++ usage pattern: `rocfft_plan` is the expensive, reusable, thread-safe object; `rocfft_execution_info` is a lightweight per-call envelope carrying runtime context (stream, work buffer). The `plan.stream` and `plan.execution_info` fields are removed from both plan structs accordingly.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -43,7 +43,7 @@ steps:
       rocmgpu: "*"
     if: build.message !~ /\[skip tests\]/
     command: "julia --project -e 'using Pkg; Pkg.update()'"
-    timeout_in_minutes: 60
+    timeout_in_minutes: 90
     env:
       JULIA_NUM_THREADS: 4
       JULIA_AMDGPU_CORE_MUST_LOAD: "1"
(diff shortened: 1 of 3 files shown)


luraess:
cscs-ci run

luraess:
cscs-ci run

luraess:
EDIT: Improved
- per-execution `rocfft_execution_info` lifetime with `try/finally`
- pointer buffers allocated before the thread-sensitive execution-info setup
- async FFT regression test repeated with an explicit `yield()` inside the spawned task

PR #936, opened 2026-06-23
Add precompilation workload and move SpecialFunctions to an extension

gbaraldi:
Mirrors CUDA.jl's startup strategy to cut time-to-first-kernel and `using` cost.

## Precompilation workload

Adds `src/precompile.jl` (included at the end of `AMDGPU.jl`) with a
PrecompileTools `@compile_workload` that runs the GPUCompiler → AMDGPU codegen
pipeline on a dummy kernel during precompilation, so the first real kernel
launch no longer JIT-compiles the entire compiler.

Measured on Julia 1.12.6 (gfx, single GPU):

| metric | before | after |
| --- | --- | --- |
| first kernel (cold) | ~8.2 s | ~1.0 s |
| first `A * B` | full first-use compile | 0 methods compiled |

Details:
- The workload builds the compiler config manually (baseline `gfx900` target), so
  it needs **neither a GPU nor ROCm discovery** — it only uses LLVM. Guarded by
  `:AMDGPU in LLVM.backends()` and `VERSION >= v"1.12-"` (matching CUDA.jl's
  foreign-MI-during-precompile workaround).
- Because the workload compiles *before* `__init__`/ROCm discovery runs,
  `libdevice_libs` is empty, which would poison the `DEVICE_LIBS` /
  `_global_hostcalls` caches with empty entries and bake them into the image
  (breaking `__ocml_*` linking at runtime). They are `empty!`-ed after the
  workload so they repopulate correctly once discovery has run.
- Adds the same explicit launch/library `precompile` directives CUDA.jl uses:
  `hipfunction`/`actual_compilation`/`hiplink`, plus the hot entry points of the
  bundled libraries (rocBLAS `gemm!`/`*`/`mul!`, rocSOLVER factorizations, rocFFT
  plans, `rand`).

## SpecialFunctions extension

Moves `SpecialFunctions` from a hard dependency to a weakdep +
`AMDGPUSpecialFunctionsExt`, extracting its device overrides out of
`device/gcn/math.jl` (the Base-math overrides stay in the package). `using
AMDGPU` no longer loads SpecialFunctions and its dependency tree; the device
overrides are registered when a user loads SpecialFunctions, same as CUDA.jl.

Verified: Base-math kernels (`sqrt`/`exp`/`sin`/`pow`), matmul, reductions, and
the extension's `erf`/`loggamma` device overrides all produce correct results.


luraess (approved):

luraess:
Thanks! Feel free to merge when ready.

luraess:
@gbaraldi is this ready?

gbaraldi:
I think so

luraess:
Anything special needed wrt rocSPARSE (in a similar fashion to rocRAND, rocBLAS)?

luraess:
Thanks!

PR #941, opened 2026-07-01
Fix segfault in qr(dA) \ b

luraess:
### Problem

On Julia 1.10+, `adjoint(Q::AbstractQ)` returns `LinearAlgebra.AdjointQ`, not `Adjoint`, making the six existing `lmul!`/`rmul!` overloads on `Adjoint{T,<:QRPackedQ}` / `Transpose{T,<:QRPackedQ}` dead code. Julia 1.12 then added:

```julia
lmul!(adjQ::AdjointQ{<:Any,<:QRPackedQ{T,<:StridedMatrix}}, B::StridedVecOrMat{T})
```

Since `ROCArray <: StridedArray`, this matches GPU arrays and calls CPU OpenBLAS with a GPU pointer → segfault.

`qr(dA) \ b` was also broken independently: stdlib's `\(Factorization, b)` allocates the output with `zeros(T, n)` — always a CPU array — then passes it through `ldiv!` mixing GPU factors with CPU data → segfault.

### Fix

- Replace the six dead `Adjoint`/`Transpose`-based overloads with `AdjointQ`-based ones, widening `B` to `StridedROCVecOrMat` to also cover SubArray views produced by stdlib internals.
- Add `Base.:\` overloads for `QR{T,<:ROCMatrix,<:ROCVector}` that call `ldiv!(F, copy(b))` directly, bypassing the `zeros(T, n)` CPU allocation.
- Widen `ormqr!` / `LAPACK.ormqr!` to accept `StridedROCVecOrMat`.

### Tests

- `qr(dA) \ b` and `qr(dA) \ B` (tall matrix, vector and matrix RHS)
- Explicit `lmul!`/`rmul!` for `Q`, `Q'`, and `transpose(Q)` (real types)
- `ldiv!(qr(dA), dx)` and `ldiv!(dy, qr(dA), dx)`

### DIFF
--- a/src/solver/highlevel.jl
+++ b/src/solver/highlevel.jl
@@ -561,23 +561,23 @@ LinearAlgebra.lmul!(
 ) where T =
     ormqr!('L', 'N', A.factors, A.τ, B)
 
+# Julia 1.10+ returns AdjointQ (not Adjoint/Transpose) from adjoint(::AbstractQ) and
+# transpose(::AbstractQ{<:Real}).  Julia 1.12 added a BLAS-backed
+# lmul!(::AdjointQ{…,<:QRPackedQ{T,<:StridedMatrix}}, ::StridedVecOrMat{T}) that
+# matches ROCArrays (ROCArray <: DenseArray <: StridedArray) and calls CPU OpenBLAS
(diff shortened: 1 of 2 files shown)


mfherbst:
Awesome, thanks @luraess !

PR #945, opened 2026-07-03
Add eigensolver support for Hermitian and Symmetric

### LINKED ISSUE #903: Generalized eigensolver for Hermitian matrices fails
Calling the generalized eigensolver `eigen(Hermitian(A), Hermitian(B))` fails when both A and B are `Hermitian{T, <:ROCArray}`. The error message reads:

```
ERROR: LoadError: PosDefException: matrix is not positive definite; Factorization failed.
```

Here is a reproducer:
```
using LinearAlgebra
using AMDGPU

### On the CPU
A = randn(50, 50)
A = A * A'
B = randn(50, 50)
B = B * B'

ref_vals, ref_vecs = eigen(A, B)
ref_hvals, ref_hvecs = eigen(Hermitian(A), Hermitian(B))

### On the GPU
A_gpu = ROCArray(A)
B_gpu = ROCArray(B)

gpu_vals, gpu_vecs = eigen(A_gpu, B_gpu) # This works, and is correct
@assert maximum(abs, ref_vals .- Array(gpu_vals)) < 1.0e-14
@assert maximum(abs, ref_vecs .- Array(gpu_vecs)) < 1.0e-14

eigen(Hermitian(A_gpu), Hermitian(B_gpu)) # This fails
```

Interestingly enough, `eigen(A, B)` without the explicit Hermitian wrapper works, despite `methods(eigen)` showing no specific method for `eigen(A::ROCArray{T, 2}, B::ROCArray{T, 2})`. I guess we are lucky that the default LinearAlgebra path works on AMD GPUs.

For completeness, I have a workaround:
```
function LinearAlgebra.eigen(A::Hermitian{T, <:ROCArray}, B::Hermitian{T, <:ROCArray}) where {T}     
    C = cholesky(B)                                                                                  
    D = C.L \ A / C.L'                                                                               
    F = eigen(Hermitian(D))                                                                          
    return (; values = F.values, vectors = C.L' \ F.vectors)                                         
end
```

luraess:
Add ROCMatrix-specific generalized eigensolver support for Hermitian and Symmetric inputs.

This makes calls like `eigen(Hermitian(A_gpu), Hermitian(B_gpu))` dispatch to rocSOLVER instead of falling through to the generic `LinearAlgebra` path.

## Changes

- Add high-level `sygvd!` / `hegvd!` rocSOLVER wrappers.
- Add `eigen(A, B)` methods for `Symmetric` and `Hermitian` ROC matrices.
- Respect wrapper `uplo`, matching Base `LinearAlgebra` behavior.
- Map positive solver `info` to `LinearAlgebra.PosDefException`.
- Added the mixed real wrapper support and tests.

## Testing

Added regression coverage for generalized `Symmetric` and `Hermitian` ROCMatrix eigensolver calls.

Fixes #903.

### DIFF
--- a/src/solver/highlevel.jl
+++ b/src/solver/highlevel.jl
@@ -787,6 +787,14 @@ function LinearAlgebra.eigen(A::Hermitian{T,<:ROCMatrix}, B::Hermitian{T,<:ROCMa
     eigen(Symmetric(A), Symmetric(B))
 end
 
+function LinearAlgebra.eigen(A::Hermitian{T,<:ROCMatrix}, B::Symmetric{T,<:ROCMatrix}) where {T<:BlasReal}
+    eigen(Symmetric(A), B)
+end
+
+function LinearAlgebra.eigen(A::Symmetric{T,<:ROCMatrix}, B::Hermitian{T,<:ROCMatrix}) where {T<:BlasReal}
+    eigen(A, Symmetric(B))
+end
+
 function LinearAlgebra.eigen(A::ROCMatrix{T}) where {T<:BlasReal}
     A2 = copy(A)
     issymmetric(A) ? Eigen(syevd!('U', A2)...) : error("GPU eigensolver supports only Hermitian or Symmetric matrices.")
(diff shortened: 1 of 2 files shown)


luraess:
Note that mixed Hermitian/Symmetric support is intentionally real-only, because complex Symmetric is not a Hermitian definite problem.

PR #949, opened 2026-07-04
Enable Dependabot for Julia dependencies

vchuravy:
Dependabot gained [native Julia support in December 2025](https://github.blog/changelog/2025-12-16-dependabot-version-updates-now-support-julia/) and is the [recommended replacement for CompatHelper](https://discourse.julialang.org/t/psa-github-dependabot-now-supports-julia/134997): its PRs trigger CI automatically (no manual close/reopen), and it also updates the checked-in `Manifest.toml`, surfacing resolver errors in the PR body when resolution fails.

This adds a `julia` ecosystem entry (weekly, ungrouped so each bump gets its own PR, matching CompatHelper's behavior) alongside the existing `github-actions` entry, and removes `.github/workflows/CompatHelper.yml` since Dependabot supersedes it.

Note: only the root environment is covered — the PSA advises against listing multiple directories (`docs/`, `test/`, `gen/`) due to resolver conflicts, recommending Julia 1.12+ workspaces instead, which isn't an option while we support 1.10.


vchuravy:
x-ref: https://github.com/trixi-framework/Trixi.jl/pull/2833#issuecomment-4881454208

PR #950, opened 2026-07-04
Allow GPUToolbox v2 and v3

vchuravy:
Bumps the GPUToolbox compat to allow v2 and v3.

The breaking changes in GPUToolbox v2/v3 are confined to the `@memoize`/`LazyInitialized` redesign (memoization and threading). AMDGPU only uses `@gcsafe_ccall`, which is unchanged through v3.0.0.

Verified that AMDGPU resolves and precompiles against GPUToolbox v3.0.0 in a clean environment.


vchuravy:
x-ref: https://github.com/trixi-framework/Trixi.jl/pull/2833#issuecomment-4880903784

PR #961, opened 2026-07-06
Switch to CompilerCaching.jl.

maleadt:


vchuravy (approved):

maleadt:
Done on my end. I'll leave it to AMDGPU.jl maintainers to merge this.

luraess:
@wsmoses we have this now failing for Enzyme tests as this PR requires now GPUCompiler v2 and Enzyme is still at 1.6.2 (see https://buildkite.com/julialang/amdgpu-dot-jl/builds/3886#019f5b1f-0fd3-4bc7-8c10-878504fd0dc9/L669). Could we bump compat and patch Enzyme?

wsmoses:
@vchuravy has a pr starting to look into it.

That said, @maleadt as the changes here don't look huge, is it possible to conditionally gate this with something like

if gpucompilerver >= 2
 a()
 else
  b()
 end
 
 that way downstream packages that haven't adapted yet [like enzyme/reactant] can still run ci with?
 
 And the same for cuda/metal?

maleadt:
I'd rather not. I've added enough cruft already in GPUCompiler just to keep supporting 1.10 for Enzyme, and I don't want that leak into the back-ends too.

wsmoses:
Fair enough, in which case would you be able to help us adapt enzyme/reactant to gpucompiler 2?

maleadt:
Valentin created PRs already. The changes are not tricky; point an LLM at my changes and let it update those PRs.

luraess:
@wsmoses let me know if you can things rolling - as ideally I'd like to have Enzyme CI passing before merging this.

wsmoses:
there's a draft pr open for enzyme by @vchuravy and no pr at all for reactant. 

without some assistance from gpucompiler folks, there's likely going to be non-trivial latency to adapt to the api churn as we're still working through some higher priority blockers atm.

christiangnrd:
@wsmoses I've opened https://github.com/EnzymeAD/Enzyme.jl/pull/3354 targeting the draft PR which keeps support for GPUCompiler v1. This doesn't seem like too much to maintain at least until most things have transitioned to GPUCompiler 2.

I gave Reactant.jl a quick look and I don't think any of the GPUCompiler methods used by Reactant.jl need to adjust to a need signature/name. Obviously behaviour may have changed and require fixing but it's possible it'll just be a compat bump

luraess:
What's the status of https://github.com/EnzymeAD/Enzyme.jl/pull/3354 ? In the meantime, I am thinking of allowing for soft fail (as CUDA does) and merge this PR.

luraess:
Thanks!

PR #962, opened 2026-07-15
Add device-pointer-mode rocBLAS level-1 reductions (nrm2!/dot!/dotc!/dotu!)

ipasichnyk:
Adds device-pointer-mode variants of the rocBLAS level-1 reductions so their scalar
result is written into a caller-provided device buffer instead of being returned to the
host. The existing `nrm2`/`dot`/`dotc`/`dotu` wrappers use a host `Ref` and return
`result[]`, which forces a device→host copy and synchronizes the stream on every call.
The new `!` variants keep the result on the GPU, enabling fully device-resident norm/dot
accumulation with no per-call sync.
## What's added
- `with_device_pointer_mode(f)` — runs `f(handle)` with `rocblas_set_pointer_mode(device)`
  and restores host mode in a `finally` (so subsequent GEMMs with host alpha/beta are
  unaffected).
- `nrm2!(result, n, X, incx)` — for `Float32`/`Float64`/`ComplexF32`/`ComplexF64`
  (`result` is the real element type).
- `dot!`/`dotc!`/`dotu!(result, n, X, incx, Y, incy)` — real `dot!` (incl. `Float16`) and
  the conjugated/unconjugated complex variants.
Each writes into a 1-element `ROCArray` `result` and returns it.


luraess:
Hi! Should this be discarded given you closed it?

If not, thanks for the addition. In that case, could you please rebase your PR commits on latest main to have a clean history.

ipasichnyk:
Yes, please discard it for a time being

PR #1004, opened 2026-07-29
don't return reshaped array from `mapreducedim!`

simeonschaub:
Same as https://github.com/JuliaGPU/CUDA.jl/pull/3219, discovered in https://github.com/JuliaGPU/GPUArrays.jl/pull/754.


luraess:
Thanks!

PR #1005, opened 2026-07-29
use AMDGPU_LLVM_Backend_jll instead of in-tree backend

### LINKED ISSUE #1002: Int128 regression in Julia 1.12
I don't have a AMD gpu that I can debug this on but something has broken with Int128 arrays in Julia 1.12. 
This was found when testing a GPUArrays.jl PR that adds Int128 to the list of eltypes tested in the testsuite: https://buildkite.com/julialang/gpuarrays-dot-jl/builds/1828/canvas?sid=019f99ef-2d3f-4354-b5aa-41128d35a9f0

simeonschaub:
We might eventually want to switch to using the clang shipped by ROCm, so that the compiler, ld.lld, device libs, and the other vendor libs always stay in sync with upstream, but let's try the jll approach for now.

fixes #1002
cc @vchuravy since you worked on JuliaGPU/GPUCompiler.jl#857


vchuravy (approved):

vchuravy:
Oh did I never finish that part :)

luraess:
Interestingly, we have documentation and benchmark CI jobs that are failing, while others pass (with exception of this failing for known reason)

simeonschaub:
Ah, I think I see what's happening: LLD_jll is a Julia stdlib, so we're forced to use the version Julia ships with. That version is too old to handle the code produced by AMDGPU_LLVM_Backend_jll, which is based on LLVM 22, though.

I (hopefully) fixed the precompile issues by just skipping creating the executable altogether, but this means we are now relying on the lld shipped by ROCm and we can't fall back to LLD_jll anymore. Maybe that's ok though and we should just remove the artifact fallback path for lld?

vchuravy:
Or have LLD be coming from AMDGPU_LLVM_Backend_jll

vchuravy (approved):

luraess:
Thanks!

PR #1009, opened 2026-08-01
Reduce per-kernel allocations from 12 down to 1

pxl-th:
- Switch from `@check` to `GPUToolbox.@checked` to force-inline hip wrappers (it also generated an unchecked version needed by graph capture stuff).
- Simplify `pack_arguments` aking to [CUDA](https://github.com/JuliaGPU/CUDA.jl/pull/3196/changes).
- Switch key for kernel instances lookup.
- `@inline` hot-path functions.
 

Script:
```julia
using AMDGPU
using BenchmarkTools

function vadd!(c, a, b)
    i = workitemIdx().x + (workgroupIdx().x - 1) * workgroupDim().x
    c[i] = a[i] + b[i]
    return
end

function vlaunch(c, a, b, count::Int)
    groupsize = 256
    gridsize = cld(length(c), groupsize)
    for _ in 1:count
        @roc groupsize=groupsize gridsize=gridsize vadd!(c, a, b)
    end
    return
end

function main()
    n = 1024
    a = AMDGPU.ones(Int, n)
    b = AMDGPU.ones(Int, n)
    c = AMDGPU.zeros(Int, n)

    vlaunch(c, a, b, 1)

    t = @benchmark vlaunch($c, $a, $b, 1000) teardown=(AMDGPU.synchronize()) evals=1 samples=1_000
    display(t); println()
    return
end
main()
```

Before:
```julia
BenchmarkTools.Trial: 1000 samples with 1 evaluation per sample.
 Range (min … max):  2.029 ms …   5.289 ms  ┊ GC (min … max): 0.00% … 55.98%
 Time  (median):     2.114 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.157 ms ± 234.436 μs  ┊ GC (mean ± σ):  1.48% ±  5.79%

  ▃▅▇█▇▅     ▁▁                                                
  ███████▇▄▇███▅▅▅▄▁▁▄▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄▅▅▁▅▅▁▄ █
  2.03 ms      Histogram: log(frequency) by time      3.45 ms <

 Memory estimate: 546.88 KiB, allocs estimate: 12000.
```

Now:
```julia
BenchmarkTools.Trial: 1000 samples with 1 evaluation per sample.
 Range (min … max):  1.565 ms …  2.656 ms  ┊ GC (min … max): 0.00% … 31.27%
 Time  (median):     1.626 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.639 ms ± 70.892 μs  ┊ GC (mean ± σ):  0.18% ±  1.88%

           ▄▅█▇▃                                              
  ▃▄▄▆▆▆▅▆▇███████▆▇▇▆▅▆▅▄▄▄▄▃▃▂▂▃▂▂▃▂▂▁▂▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▂▁▂ ▃
  1.57 ms        Histogram: frequency by time        1.84 ms <

 Memory estimate: 109.38 KiB, allocs estimate: 1000.
```

### DIFF
--- a/src/compiler/codegen.jl
+++ b/src/compiler/codegen.jl
@@ -36,8 +36,8 @@ mutable struct HIPResults
     HIPResults() = new(nothing, nothing, Symbol[], Tuple{HIP.HIPDevice,HIP.HIPFunction}[])
 end
 
-# hash(fun, hash(f, hash(tt))) => HIPKernel
-const _kernel_instances = Dict{UInt, Runtime.HIPKernel}()
+# (objectid(source), hash(fun), f) => HIPKernel
+const _kernel_instances = Dict{Any, Any}()
 
 GPUCompiler.runtime_module(@nospecialize(::HIPCompilerJob)) = AMDGPU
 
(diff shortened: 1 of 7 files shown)


luraess (approved):

luraess:
Awesome - thanks!

PR #1013, opened 2026-08-10
Let some wrappers use strided arrays

kshyatt:
Right now they only accept stride=1 `ROCMatrix` in many places, would be great to allow `StridedROCMatrix` analogous to CUDA where possible.

### DIFF
--- a/src/solver/highlevel.jl
+++ b/src/solver/highlevel.jl
@@ -29,7 +29,7 @@ for (fname, elty) in (
     (:rocsolver_zpotrs, :ComplexF64),
 )
     @eval begin
-        function potrs!(uplo::Char, A::ROCMatrix{$elty}, B::StridedROCVecOrMat{$elty})
+        function potrs!(uplo::Char, A::StridedROCMatrix{$elty}, B::StridedROCVecOrMat{$elty})
             chkuplo(uplo)
             n = checksquare(A)
             m = size(B, 1)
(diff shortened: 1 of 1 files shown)


simeonschaub:
Probably needs some tests with non-unit stride to make sure we don't run into any correctness issues. Does CUDA already have tests for this?

kshyatt:
I think [not](https://github.com/JuliaGPU/CUDA.jl/blob/main/lib/cusolver/test/dense/qr.jl), it just lets you pass strided arrays in and YOLO

luraess:
Cool - thanks for the addition! And yes, if possible we could be on the safer side to have some test of striding...

kshyatt:
I'll try to add some tests :)

simeonschaub:
Hmm, `stride(A, 1)` is never actually passed, I don't think these functions accept a non-1 stride in the first dimension. So `ROCMatrix` here seems correct, it looks like CUDA is too lenient and will return wrong results

kshyatt:
But isn't something with a `stride(A, 1) = 1` and `stride(A, 2) = 2)` still a `StridedArray`?

simeonschaub:
Ah, I see what you mean! I agree that's useful, should we add a check that `stride(A, 1) == 1`? Though I guess if CUDA doesn't check either we can just follow its example for now and track this in a separate issue

luraess (approved):
Thanks! Up to you if you'd like to add any test.

kshyatt:
OK I may revisit this in the future for more tests, in that case

PR #1024, opened 2026-08-15
refactor discovery, ship therock tarballs as artifacts

simeonschaub:
Some of the AI comments still need to be cleaned up but the logic looks reasonable from a first review, so I thought I'd put this up as a draft for feedback.

It turns out MIOpen needs all of libllvm and libclang, so unlike discussed in https://github.com/JuliaGPU/AMDGPU.jl/pull/919#issuecomment-5278973837, I went back to shipping therock artifacts directly as done in that PR, since we don't gain much by repackaging anymore.

Otherwise, do a split similar to CUDA.jl of having ROCm_Runtime be a pseudo-jll providing automatic downloads through artifacts and having manual discovery live in ROCm_Runtime_Discovery. Either one of those is then loaded depending on a preference


luraess:
Cool!

Do we also handle Instinct GPUs? I see they seem omitted in explicit cases which though relate mostly to Windows OS from the comments.

simeonschaub:
Yes, instinct should work just fine on Linux, it's just Windows that doesn't support them because upstream doesn't provide tarballs for that combination.

We have external CI on MI300, correct? Could you trigger a run, or can I even do it myself now?

luraess:
I see, the marketing name is Windows only indeed, so all seems fine and Instinct GPUs are included. Just tested this locally and all works fine.

luraess:
> We have external CI on MI300, correct? Could you trigger a run, or can I even do it myself now?

This would only be possible if the PR comes from an AMDGPU branch (limitation from external CI). Is there a way we could switch to a AMDGPU branch now that you have write access?

simeonschaub:
Hmm, not sure there is a good way to do this with GitHub except for just opening a new PR

PR #1027, opened 2026-08-17
fix nightly failures

simeonschaub:
Previously, the `noinline` attribute emitted by Julia for some error paths collided with us setting `alwaysinline`. Make the `alwaysinline` take precedence here


luraess:
Nightly is allow to soft fail, which still happens here. Is something else missing?

simeonschaub:
Yes, seems like the fix just surfaced another issue

simeonschaub:
Should be addressed by https://github.com/JuliaLang/julia/pull/62807

simeonschaub:
It looks like the nightly timeout might need to be increased? (Nightly is still expected to fail until https://github.com/JuliaLang/julia/pull/62823 is merged)

luraess:
Yes, good catch. One could bump it to 90 as for the other tests.

PR #1028, opened 2026-08-17
refactor discovery, ship therock tarballs as artifacts (2)

simeonschaub:
(Same as #1024, reopened from the upstream branch to run MI300 CI)

Some of the AI comments still need to be cleaned up but the logic looks reasonable from a first review, so I thought I'd put this up as a draft for feedback.

It turns out MIOpen needs all of libllvm and libclang, so unlike discussed in https://github.com/JuliaGPU/AMDGPU.jl/pull/919#issuecomment-5278973837, I went back to shipping therock artifacts directly as done in that PR, since we don't gain much by repackaging anymore.

Otherwise, do a split similar to CUDA.jl of having ROCm_Runtime be a pseudo-jll providing automatic downloads through artifacts and having manual discovery live in ROCm_Runtime_Discovery. Either one of those is then loaded depending on a preference

PR #1031, opened 2026-08-17
CSCS MI300 CI: select the ROCm provider explicitly

luraess:
Follow-up to #1028 for the CSCS MI300 job, which currently fails every test.

`ci/cscs-mi300.yml` used `JULIA_AMDGPU_DISABLE_ARTIFACTS=1` to pick up the uenv's ROCm. That variable no longer exists after the discovery refactor, so the job silently resolved to the artifacts instead, and every test died with `hipErrorOutOfMemory` out of `hipStreamCreateWithPriority` with no `hipMalloc` ever failing.

This selects the provider through `ROCm_Runtime`'s `local` preference, set before `Pkg.instantiate()` since it also gates artifact resolution, and devs the two workspace packages the way `.buildkite/pipeline.yml` now does. The resolved provider is asserted against the requested one, so a change of default can't silently swap the ROCm underneath the job again.

The script is parameterised by `ROCM_LOCAL` so both providers can be run, but only the uenv variant is enabled. The artifact path can't currently get off the ground on gfx942, see https://github.com/ROCm/TheRock/issues/7426, That job is left commented out, to be re-enabled (with `allow_failure`) once that is resolved.

### DIFF
--- a/ci/cscs-mi300.yml
+++ b/ci/cscs-mi300.yml
@@ -73,12 +73,12 @@ UnitTest julia 1.12 (system ROCm):
   script: *unit_test_script
 
 # Coverage of the artifact path -- what AMDGPU.jl uses by default -- is worth
-# having here, since this is the only gfx942 CI in the project. It is disabled
-# for now because it cannot get off the ground on this machine: TheRock ships a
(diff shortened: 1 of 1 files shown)


luraess:
cscs-ci run

PR #1032, opened 2026-08-17
Fix generic_matmul handling of size zero arrays

kshyatt:
Right now the `rmul!` *forces* `beta = 0` even when that should not be the case. However, the output should still be updated even if the input array is size 0.

### DIFF
--- a/test/hip_rocarray/blas.jl
+++ b/test/hip_rocarray/blas.jl
@@ -816,8 +816,8 @@ end
         end
         # matrix * vector 
         C0 = rand(T, m)
-        A0 = rand(T, k)
-        B0 = rand(T, k, m)
+        A0 = rand(T, m, k)
+        B0 = rand(T, k)
         for (α, β) in ( (-one(T), one(T)), (one(T), 2*one(T)), (one(T), zero(T))) 
             Ccpu = copy(C0)
             mul!(Ccpu, A0, B0, α, β)
--- a/test/hip_rocarray/blas.jl
+++ b/test/hip_rocarray/blas.jl
@@ -790,10 +790,34 @@ end
 
 @testset "Handling size zero arrays correctly" begin
     for T in (Float32, Float64, ComplexF32, ComplexF64)
-        m, k, n = (12, 6, 0)
+        m, n, k = (12, 6, 0)
+        # matrix * matrix
         C0 = rand(T, m, n)
         A0 = rand(T, m, k)
         B0 = rand(T, k, n)
+        for (α, β) in ( (-one(T), one(T)), (one(T), 2*one(T)), (one(T), zero(T))) 
+            Ccpu = copy(C0)
+            mul!(Ccpu, A0, B0, α, β)
+            Cgpu = ROCArray(copy(C0))
+            mul!(Cgpu, ROCArray(A0), ROCArray(B0), α, β)
+            @test Ccpu ≈ Array(Cgpu)
+        end
+        # C0 is NaN and beta is zero to test beta is applied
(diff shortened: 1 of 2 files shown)


test/hip_rocarray/blas.jl
```diff
@@ -786,6 +786,22 @@ end
             end
         end
     end
+end
 
+@testset "Handling size zero arrays correctly" begin
+    for T in (Float32, Float64, ComplexF32, ComplexF64)
+        m, k, n = (12, 6, 0)
```
simeonschaub: Shouldn't it be k = 0 instead of n = 0 here? Otherwise, C0 is just empty and this doesn't make a difference, right?

simeonschaub (commented):
Thanks! Could you add a test for the case where both `C0` and `A0` are vectors?

Maybe we can also add a test where `C0` is filled with `NaN` and beta is zero to make sure beta is a strong zero? I assume that's why you used `_rmul_or_fill!`

test/hip_rocarray/blas.jl
kshyatt: Ah yes good catch

test/hip_rocarray/blas.jl
```diff
@@ -786,6 +786,46 @@ end
...
+            Ccpu = copy(C0)
+            mul!(Ccpu, A0, B0, α, β)
+            Cgpu = ROCArray(copy(C0))
+            mul!(Cgpu, ROCArray(A0), ROCArray(B0), α, β)
+            @test Ccpu ≈ Array(Cgpu)
+        end
+        # matrix * vector 
+        C0 = rand(T, m)
+        A0 = rand(T, k)
+        B0 = rand(T, k, m)
```
simeonschaub: Should this be
```suggestion
        A0 = rand(T, m, k)
        B0 = rand(T, k)
```
?

test/hip_rocarray/blas.jl
```diff
@@ -786,6 +786,46 @@ end
...
+        for (α, β) in ( (-one(T), one(T)), (one(T), 2*one(T)), (one(T), zero(T))) 
+            Ccpu = copy(C0)
+            mul!(Ccpu, A0, B0, α, β)
+            Cgpu = ROCArray(copy(C0))
+            mul!(Cgpu, ROCArray(A0), ROCArray(B0), α, β)
+            @test Ccpu ≈ Array(Cgpu)
+        end
+        # C0 is NaN and beta is zero to test beta is applied
+        C0 = rand(T, m, n)
+        C0[m-1, n-1] = T(NaN)
```
simeonschaub:
```suggestion
        C0 = fill(T(NaN), m, n)
```

kshyatt: Isn't C0 still CPU side at this point?

test/hip_rocarray/blas.jl
kshyatt: You asked above for the case where C0 and A0 are vectors?

simeonschaub: Ah, sorry, that's my bad! I meant C0 and B0 in order to cover the gemv method

test/hip_rocarray/blas.jl
simeonschaub: Oops, yes! I updated my suggestion

simeonschaub (approved):
Looks great, thanks!

luraess (approved):

kshyatt:
Hmmm this passed for me locally...

luraess:
Thanks!

PR #1037, opened 2026-08-18
Relocate generated code via patch strategy

pxl-th:
- Relocate generated code via patch strategy.
- Link device libraries only for the toplevel LLVM module. Others don't need it and it's wasteful.
- On-demand & lazy device library loading.
- Add memset and use it for fill.
- Use single LLVM context for filtering out ghost types.


luraess:
cscs-ci run

maleadt:
Would be good to try and actually exercise this by precompiling some native code.

pxl-th:
> Would be good to try and actually exercise this by precompiling some native code.

Yes, I'll add it in a bit along with tests.

PR #1041, opened 2026-08-19
Override `sind`/`cosd` to avoid the malloc hostcall

luraess:
Kernels calling `sind`, `cosd`, `tand`, `sincosd`, `secd`, `cscd` or `cotd` report `Global hostcalls detected! - Hostcalls: [:malloc_hostcall]` and spin up its persistent host-side polling task, without allocating anything.

Base's `sind`/`cosd` construct `DomainError(x, msg)` inline, and `DomainError.val` is untyped, so the argument gets boxed: on device that lowers to `gpu_malloc` then `Device.malloc`, writing the `__malloc_hostcall` marker global on a branch that never executes.

Fix: `@device_override` for `Base.Math.sind` and `Base.Math.cosd` with `@gpu_throw`. Base defines the other five in terms of those two, so the overlay covers all of them. The bodies are copied from Base verbatim apart from the throw (the algorithm is identical across Julia 1.10 to 1.13); `sin(deg2rad(x))` would be shorter but loses the exact zeros, `sind(180f0)` giving `-8.7e-8`, and would silently diverge from CPU results.

Tested on gfx1100: no `__malloc_hostcall` in the IR for all seven functions, results bit-identical to the CPU over 22 angles in Float32 and Float64, and `sind(Inf)` still raises `GPU Kernel Exception: DomainError`.

Added new "Degree trigonometry" testset in `test/device/math.jl`.

### DIFF
--- a/src/device/quirks.jl
+++ b/src/device/quirks.jl
@@ -41,6 +41,65 @@ end
 @device_override Base.Math.sincos_domain_error(x) =
     @gpu_throw "DomainError: sincos(x) is only defined for finite x"
 
+# Bodies copied from Base (`base/special/trig.jl`) with the inline `DomainError`
+# throw replaced: boxing its untyped `val` field emits a device-side allocation.
+@device_override function Base.Math.sind(x::Real)
+    if isinf(x)
+        @gpu_throw "DomainError: sind(x) is only defined for finite x"
+    elseif isnan(x)
+        return x
+    end
+
+    rx = copysign(float(rem(x, 360)), x)
+    arx = abs(rx)
+
+    if rx == zero(rx)
+        return rx
+    elseif arx < oftype(rx, 45)
+        return Base.Math.sin_kernel(Base.Math.deg2rad_ext(rx))
+    elseif arx <= oftype(rx, 135)
+        y = Base.Math.deg2rad_ext(oftype(rx, 90) - arx)
+        return copysign(Base.Math.cos_kernel(y), rx)
+    elseif arx == oftype(rx, 180)
+        return copysign(zero(rx), rx)
+    elseif arx < oftype(rx, 225)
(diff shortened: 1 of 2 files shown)


simeonschaub (approved):
It's unfortunate we need to duplicate so much code from base here. Could we perhaps allow passing a function name to [`sin_domain_error`](https://github.com/JuliaLang/julia/blob/ddf0c8b8a47eec6abdca05397939b28e62d9839d/base/special/trig.jl#L28) and have base use that, similar to `throw_complex_domainerror`?

That will only cover future Julia releases though, so I'm ok with this as a temporary solution for now

luraess:
Agreed, btw seems in the same file that Base already has `@noinline sincos_domain_error(x)` (`base/special/trig.jl:165`), which both AMDGPU.jl (and CUDA.jl) override. `sind`/`cosd` just never got the same treatment. Generalising it to take the function name, like `throw_complex_domainerror(f::Symbol, x)`, would let us delete both copied bodies here, and could possibly fix CUDA.jl and Metal.jl at the same time. Maybe one could open a PR for it on Base?

There may be some "issue" with compat though. AMDGPU supports Julia 1.10, so a helper landing in 1.13 leaves every currently supported version needing the copies. Then yeah - idk, maybe not good to have duplicate, on the other hand it's only `sind` and `cosd`, since all remaining are defined in terms of those two.

simeonschaub:
> Maybe one could open a PR for it on Base?

Yes, that's what I was thinking. Do you want to prepare one?

Yeah, the compat issues are why I was suggesting this as a temporary solution, but I think it's still worth fixing this in Base, so we can eventually drop it once a Julia release containing the fix becomes lts

luraess:
This would address https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1210#pullrequestreview-4981256952 as well.

PR #1044, opened 2026-08-21
use device libs from AMDGPU_LLVM_Backend_jll (2)

simeonschaub:
We downgrade these into a scratch space to be compatible with Julia's LLVM version

ref https://github.com/JuliaGPU/AMDGPU.jl/pull/919#issuecomment-5278973837

Reopened to run MI300 CI


simeonschaub:
OK, I think this should be good to go now

luraess:
Running `downgrade_device_libs` against the real JLLs on 1.10/1.11 (target 14) and 1.12/1.13 (target 18) — all 70 `.bc` files convert and re-parse, ~0.4 s one-time. A few things that may still need attention (robot helped):

1. The `LLVMDowngrader_jll = "0.9.1"` floor is undocumented. With 0.9.0, `asanrtl.bc` fails at target 14:

```
LLVM ERROR: pointers in constant aggregates are not yet supported by the IR downgrader
```

`run()` throws, which aborts the whole loop, so 0.9.0 silently drops the entire artifact path on 1.10/1.11. Worth a comment so the bound doesn't get relaxed later?

2. AMDGPU only links `hc`/`hip`/`irif`/`ockl`/`opencl`/`ocml` plus the `oclc_*` control libs. `asanrtl.bc` is in neither `ondemand_libs()` nor `control_libs()`. Skipping files that fail (or only converting the ones actually used) makes this robust against the next downgrader gap instead of dependent on a compat floor.

3. `isdefined(AMDGPU_LLVM_Backend_jll, :bitcode_path)` is doing more work than it looks. 22.1.8**+0** exports only `llc, lld`; `bitcode_path` first appears in +1. Since `compat = "22.1.8"` admits +0, that check is the only thing between a +0 resolve and a silent fallback. Is a (comment or) a tighter bound needed?

4. Quiet fallback? `from_artifact` is unconditionally `true`, so on failure the user gets a warning and then `libdevice_libs == ""` on any host without a system ROCm, surfacing much later as `unsupported call to __ocml_*`. Also: `@get_scratch!` writes to the depot from `__init__`, so a **read-only shared depot** (common on HPC, relevant to the CSCS MI300 target) fails on every load. Maybe have the warning name the consequence when `find_device_libs` also comes back empty.

Additional: `v"5"`/`v"7"` are unreachable (floor is 1.10 = LLVM 15) and fail anyway on `llvm.amdgcn.implicitarg.ptr`; `maximum` would throw on an empty iterator if `libllvm_version < 5`. And the marker + per-file `mv` leaves a small window on 1.10/1.11, where `mv(force=true)` does `rm(dst)` then `rename` rather than renaming straight over (fine on >= 1.12). A `Pidfile.mkpidlock` could close it.

luraess (approved):
Seems mostly good. I left a few comments while checking things also with the robot's help.

simeonschaub:
Addressed points 2 and 4 as well as the one regarding the LLVM 5 and 7 paths being unreachable.

Re 1, I think just the compat entry is enough, I don't know why we would ever accidentally relax it. I think 3 is fine too, since we just fall back to the old behavior if the user for some reason is still on the old backend jll. I don't think Pkg understands compat entries depending on the build number

luraess:
Agree. Thanks!

PR #1045, opened 2026-08-21
[CI] Improvements

christiangnrd:
Switch to test matrices so it's easier to make changes. Also set `--quickfail` on nightly

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -31,10 +31,25 @@ steps:
     if: build.message !~ /\[skip docs\]/
     timeout_in_minutes: 20
 
-  - label: "Julia 1.10"
+  - label: "Julia {{matrix.version}}"
+    matrix:
+      setup:
+        version:
+          - "1.10"
+          - "1.11"
+          - "1.12"
+          - "1.13"
+          - "nightly"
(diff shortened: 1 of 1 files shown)


christiangnrd:
I cancelled the jobs that this PR doesn't touch for runner resource reasons

pxl-th:
I'd still go with removing 1.11 from CI and making it best-effort for now, waiting ~8 hours for PR to run in CI is a bit hilarious.
But I saw @simeonschaub [does something](https://github.com/JuliaGPU/AMDGPU.jl/pull/1044/commits/aaf721c635ff4fe7d7a1d89b71908cd26bb04d60) with 1.11, so I won't push if it's really needed.

simeonschaub:
I'm not opposed to dropping 1.11 support to just be on a best-effort basis and not run CI. I know a lot of the Enzyme ecosystem is still kind of stuck at 1.11, but that might be changing, so the question is really whether there's still a noteworthy userbase there, as it's likely 1.11 will regress with no CI

pxl-th:
https://github.com/JuliaGPU/AMDGPU.jl/pull/1037 made CI ~2x faster, so let's leave 1.11 :)

PR #1049, opened 2026-08-24
rocBLAS: route 16-bit mul! through gemm_ex (Float16 3→190 TFLOPS on MI300A)

### LINKED ISSUE #1048: Float16 mul! falls through to GPUArrays' generic kernel (~65x slower than rocBLAS gemm_ex)
Float16 `mul!` currently reaches neither rocBLAS nor any tuned kernel — it falls through to GPUArrays' generic matmul. Measured on an MI300A (gfx942, Julia 1.12.6; identical numbers under ROCm 6.4.2, 7.2.4 and 7.14.0), a 4096³ f16 GEMM runs at **3.0 TFLOPS**, when rocBLAS's tuned path delivers ~200 on the same machine.

Two stacked causes:

**1. Dispatch fallthrough.** The rocBLAS branch in `LinearAlgebra.generic_matmatmul!` (src/blas/highlevel.jl) requires `T <: ROCBLASFloat`, which excludes `Float16` (and `BFloat16`). On Julia ≥ 1.12 the `generic_matmatmul_wrapper!` overload is declared for `T <: ROCBLASFloatWithHalf` — Float16 included — but it forwards to that same `generic_matmatmul!`, so Float16 falls to `GPUArrays.generic_matmatmul!` anyway. Julia 1.11 hits the same eltype branch, so the fallthrough exists there too.

**2. The typed entry point would be the wrong target anyway.** `rocBLAS.gemm!` maps Float16 to legacy `rocblas_hgemm`, which Tensile doesn't tune. The tuned path is `rocblas_gemm_ex` with f16 operands and an f32 compute type — which is also the numerically better route (f32 accumulation). AMDGPU.jl carries only the raw generated binding (`rocblas_gemm_ex_64`), with no high-level wrapper and nothing dispatching to it.

Measured ladder, 4096³ f16 (min of 20, MI300A):

| path | TFLOPS |
|---|---|
| `mul!` today (GPUArrays fallback) | 3.0 |
| `rocBLAS.gemm!` → `rocblas_hgemm` | 69.6–71.8 |
| `rocblas_gemm_ex`, f16 in / f32 compute | 189–224 |

For f32/f64, `gemm_ex` and the typed `gemm!` perform identically (50.2 vs 50.2 and 52.2 vs 52.7 TFLOPS respectively), so routing through `gemm_ex` carries no risk for the types that already work.

Repro:

```julia
using AMDGPU, LinearAlgebra
A, B = ROCArray(rand(Float16, 4096, 4096)), ROCArray(rand(Float16, 4096, 4096))
C = similar(A)
mul!(C, A, B)  # ~85 ms on MI300A; ~1.4 ms via rocblas_gemm_ex
```

PR incoming: a `rocBLAS.gemm_ex!` wrapper (mirroring `CUBLAS.gemmEx!`) plus dispatching `Float16`/`BFloat16` `mul!` through it with Float32 accumulation — validated on MI300A at 190 (f16) / 208 (bf16) TFLOPS with the existing Level 3 tests passing.

gbaraldi:
Fixes #1048.

`Float16` `mul!` currently falls through to GPUArrays' generic kernel (the rocBLAS branch in `generic_matmatmul!` requires `T <: ROCBLASFloat`; the Julia-1.12 `generic_matmatmul_wrapper!` overload for `ROCBLASFloatWithHalf` forwards into that same branch). And the typed `gemm!` would be the wrong target anyway: Float16 there maps to legacy `rocblas_hgemm`, which Tensile doesn't tune.

This PR:

- adds `rocBLAS.gemm_ex!`, a high-level wrapper over `rocblas_gemm_ex_64` (mirrors `CUBLAS.gemmEx!`), with a datatype map for f16/bf16/f32/f64/c32/c64 and a compute-type rule giving 16-bit operands Float32 accumulation;
- dispatches `Float16`/`BFloat16` `mul!` through it in `generic_matmatmul!` (covers Julia 1.11 and 1.12 in one place);
- tests: 16-bit `mul!` against an f32 host reference, plus direct `gemm_ex!` tests for f16/bf16/f32. The existing Level 3 `mul!` testset already covers Float16 with every transpose/adjoint/alpha combination, and now exercises the new path.

Measured on an MI300A (gfx942, ROCm 7.2.4, Julia 1.12.6), 4096³, min of 20:

| eltype | `mul!` before | `mul!` after | legacy `gemm!` |
|---|---|---|---|
| Float16 | 3.0 TFLOPS | **190.3** | 69.6 |
| BFloat16 | (GPUArrays fallback) | **207.5** | n/a |
| Float32 | 49.9 | 49.9 | 50.2 |
| Float64 | 52.5 | 52.5 | 52.7 |

`gemm_ex` and typed `gemm!` are identical for f32/f64 (same Tensile kernels underneath), so this routing carries no regression risk; f32/f64 are left on their existing path regardless. Correctness validated for f16/bf16 × {N, T} at odd sizes (rel err ≤ 3.6e-4 / 2.9e-3 vs f32 host reference).

Only the ILP64 `rocblas_gemm_ex_64` binding exists in the generated wrappers, which is what's used here (rocBLAS ≥ 4.0 / ROCm ≥ 6.0).

### DIFF
--- a/src/blas/wrappers.jl
+++ b/src/blas/wrappers.jl
@@ -768,6 +768,54 @@ for (fname, elty) in
     end
 end
 
+## gemm_ex: GEMM through rocBLAS's Tensile-tuned mixed-precision path.
+# The typed gemm entry points map Float16 to legacy `rocblas_hgemm`, which is
+# not tuned (~3x slower than gemm_ex on MI300); gemm_ex with a Float32
+# compute type is the fast (and more accurate) route for 16-bit eltypes.
+
+gemmex_datatype(::Type{Float16}) = rocblas_datatype_f16_r
+gemmex_datatype(::Type{BFloat16}) = rocblas_datatype_bf16_r
+gemmex_datatype(::Type{Float32}) = rocblas_datatype_f32_r
+gemmex_datatype(::Type{Float64}) = rocblas_datatype_f64_r
+gemmex_datatype(::Type{ComplexF32}) = rocblas_datatype_f32_c
+gemmex_datatype(::Type{ComplexF64}) = rocblas_datatype_f64_c
+
+# 16-bit inputs accumulate in Float32; everything else computes in its own type
+gemmex_compute_type(TA, TB, TC) = promote_type(TA, TB, TC)
+gemmex_compute_type(::Type{Float16}, ::Type{Float16}, TC) = Float32
+gemmex_compute_type(::Type{BFloat16}, ::Type{BFloat16}, TC) = Float32
+
+function gemm_ex!(
+    transA::Char, transB::Char, alpha::Number,
+    A::StridedROCVecOrMat, B::StridedROCVecOrMat, beta::Number,
+    C::StridedROCVecOrMat;
+    compute_type::Type = gemmex_compute_type(eltype(A), eltype(B), eltype(C)),
+)
+    m = size(A, transA == 'N' ? 1 : 2)
+    k = size(A, transA == 'N' ? 2 : 1)
+    n = size(B, transB == 'N' ? 2 : 1)
+    if m != size(C, 1) || n != size(C, 2) || k != size(B, transB == 'N' ? 1 : 2)
+        throw(DimensionMismatch(""))
+    end
+
+    lda = max(1, stride(A, 2))
+    ldb = max(1, stride(B, 2))
+    ldc = max(1, stride(C, 2))
+    α = Ref(convert(compute_type, alpha))
+    β = Ref(convert(compute_type, beta))
+    (; handle, stream) = lib_state()
+    GC.@preserve A B C begin
+        rocblas_gemm_ex_64(handle, transA, transB, m, n, k,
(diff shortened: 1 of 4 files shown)


luraess (approved):
LGTM. Thanks! Maybe one could check whether some of the in-line comments could be trimmed further to not boost the loc.

simeonschaub:
I don't really understand the table. Why does it show 3 TFLOPS as the before number?

gbaraldi:
Mul! for float16 was just falling back to the generic thing

luraess:
> I don't really understand the table

All numbers are in TFlops

luraess:
Thanks!

src/blas/wrappers.jl
```diff
@@ -767,6 +767,49 @@ for (fname, elty) in
         end
     end
 end
+\
```
christiangnrd, line 770: Accidental?

luraess: Good catch

Issue #1040, opened 2026-08-19
Windows: GPU architecture detection for artifact selection is untested on real hardware


luraess:
On Linux the ROCm artifact is selected from `gfx_target_version` in the KFD topology. Windows has no KFD, so `rocm_arch()` falls back to matching GPU marketing names reported by `EnumDisplayDevicesW` against a fixed table (`ROCm_Runtime/.pkg/platform_augmentation.jl:43`):

```julia
const device_name_archs = [
    ["8050s", "8060s", "device 1586"] => "gfx1151",
    ["880m", "890m"]                  => "gfx1150",
    ["r9700", "9060", "9070"]         => "gfx120X",
    ["7700", "7800", "7900", "v710"]  => "gfx110X",
    ["6800", "6700", "6600", "6500"]  => "gfx103X",
]
```

The adapter string must also contain `radeon` or `amd`, otherwise the entry is skipped.

This path is not yet tested.

## MWE

On Windows, with any AMD GPU:

```julia
using AMDGPU
AMDGPU.versioninfo()

RR = AMDGPU.ROCm_Runtime
@show RR.windows_video_device_names()
@show RR.rocm_arch()
@show RR.is_available()
@show AMDGPU.functional()
```

Please report your GPU model and the output.

Expected: `windows_video_device_names()` lists your adapter, and `rocm_arch()` maps it to the matching `gfx` family (for example an RX 6700 XT should give `gfx103X`). An empty `rocm_arch()` or `is_available() == false` means the adapter string did not match the table and no artifact was resolved.

As a workaround in that case, the architecture can be set explicitly:

```julia
AMDGPU.set_rocm_version!(arch="gfx1031")  # substitute your own target
```

Reports of the exact adapter string are useful even when detection works.

ejmeitz:
I'm on Windows 11 with a 6700 XT. It seems to have run correctly. `AMD.versioninfo` does hang for a lot longer than I would expect, maybe 20 seconds. After compilation maybe 1-2 seconds. Not sure why `windows_video_device_names` shows that I have 7 GPUs. I only have one. 

```julia
julia> AMDGPU.versioninfo()
AMDGPU versioninfo
ROCm provider: downloaded artifacts
┌───────────┬──────────────────┬────────────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ Available │ Name             │ Version    │ Path                                                                                                                                           │
[...]

PR #1051, opened 2026-08-25
Probe rocSPARSE with `Libdl` only

luraess:
Relates to #1040. Follow-up to #1001.

#1001 moved the rocSPARSE version query into a subprocess so a segfault on a broken ROCm install (#920) degrades to `err` instead of killing the session. But the child ran `using AMDGPU` under the active project which is expensive.

The child now uses `Base.Libc.Libdl` only: it dlopens `librocsparse` by the absolute path the parent already knows and calls `rocsparse_create_handle` / `rocsparse_get_version` directly. The result is memoized for the session.

cc @simeonschaub

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1,10 +1,8 @@
-# Run `code` in a subprocess with `args` as its `ARGS`; returns stdout, or
-# `nothing` on crash, timeout, or nonzero exit. The empty `JULIA_LOAD_PATH` keeps
-# the child out of the active project, so `code` must only use `Base`.
-function _probe_subprocess(
-    code::String, args::Vector{String}; timeout::Real = 20,
-)::Union{String, Nothing}
-    cmd = `$(Base.julia_cmd()) --startup-file=no -O0 --compile=min -e $code $args`
+# Run `code` in a subprocess and return its stdout, or `nothing` on crash,
+# timeout, or nonzero exit. The empty `JULIA_LOAD_PATH` keeps the child out of
+# the active project, so `code` must only use `Base`.
+function _version_subprocess(code::String; timeout::Real = 20)::Union{String, Nothing}
+    cmd = `$(Base.julia_cmd()) --startup-file=no -O0 --compile=min -e $code`
     cmd = addenv(cmd, "JULIA_LOAD_PATH" => "")
     out = IOBuffer()
     try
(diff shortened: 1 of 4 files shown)


src/utils.jl
```diff
@@ -25,14 +26,33 @@ function _version_subprocess(code::String; timeout::Real = 60)
...
+    ccall(L.dlsym(lib, :rocsparse_create_handle), Cint,
+        (Ptr{Ptr{Cvoid}},), handle) == 0 || exit(2)
+    version = Ref{Cint}(0)
+    ccall(L.dlsym(lib, :rocsparse_get_version), Cint,
+        (Ptr{Cvoid}, Ptr{Cint}), handle[], version) == 0 || exit(2)
+    print(version[])
+    """
+
+# `nothing` until probed, then the version string or `"err"`.
+const _ROCSPARSE_VERSION = Ref{Union{Nothing, String}}(nothing)
```
simeonschaub:
```suggestion
_ROCSPARSE_VERSION::Union{Nothing, String} = nothing
```
Or even just as a plain string and we initialize with `""`?

src/utils.jl
```diff
@@ -25,14 +26,33 @@ function _version_subprocess(code::String; timeout::Real = 60)
...
+    const L = Base.Libc.Libdl
+    lib = L.dlopen(ARGS[1])
+    handle = Ref{Ptr{Cvoid}}(C_NULL)
+    ccall(L.dlsym(lib, :rocsparse_create_handle), Cint,
+        (Ptr{Ptr{Cvoid}},), handle) == 0 || exit(2)
+    version = Ref{Cint}(0)
+    ccall(L.dlsym(lib, :rocsparse_get_version), Cint,
+        (Ptr{Cvoid}, Ptr{Cint}), handle[], version) == 0 || exit(2)
+    print(version[])
+    """
```
simeonschaub: We could even interpolate the rocSPARSE lib path directly into the ccall and skip Libdl altogether?

src/utils.jl
```diff
@@ -25,14 +26,33 @@ function _version_subprocess(code::String; timeout::Real = 60)
...
-_rocsparse_version_isolated(; timeout::Real = 60) = _version_subprocess("""
-    using AMDGPU
-    AMDGPU.functional(:rocsparse) || exit(2)
-    print(AMDGPU.rocSPARSE.version())
-    """; timeout)
+# rocSPARSE's version query needs a handle, and creating one can segfault on
+# broken ROCm installs (issue #920), so run it out-of-process. `Libdl` only, so
+# the child needs no package environment.
+const _ROCSPARSE_VERSION_PROBE = """
+    const L = Base.Libc.Libdl
```
simeonschaub:
```suggestion
    using Base.Libc.Libdl: dlopen, dlsym
```
and adapt the below accordingly. AI seems to really like defining one-letter constants for some reason

simeonschaub (commented):
Yes, seems pretty much what I had in mind, thanks! Left some comments below

src/utils.jl
luraess: Implemented the plain string approach

src/utils.jl
luraess: Done as suggested.

src/utils.jl
luraess: No more Libdl

luraess:
cscs-ci run

simeonschaub (approved):

PR #1054, opened 2026-08-26
Widen dispatch packet loads

sheevy:
This is a completely vibecoded fix for the issue I saw when running my kernel.
I run it via `ParallelStencil.jl` and can run against `CUDA.jl` (on H200) and `AMDGPU.jl` (on MI355X). Performance on MI355X was surprisingly worse than on H200. This should not be the case, as my kernel is memory bandwidth bound, and MI355X is 8TB/s vs only 4.8TB/s for H200.

I let Claude loose to fix the issue, and this is what it came back with. I'm not gonna lie, I don't understand those changes. Hopefully they are meaningful i.e not slop.

### DIFF
--- a/src/device/gcn/indexing.jl
+++ b/src/device/gcn/indexing.jl
@@ -57,16 +57,40 @@ end
 
             # load the index
             offset = base + ((off - 1) * aligned_sizeof(T))
-            idx_ptr_i8 = inbounds_gep!(builder, T_int8, ptr, [ConstantInt(offset)])
-            idx_ptr_T = bitcast!(builder, idx_ptr_i8, T_ptr_T)
-            idx_T = load!(builder, T_T, idx_ptr_T)
-            idx = zext!(builder, idx_T, T_int32)
+            if aligned_sizeof(T) < 4
+                # Load the aligned dword containing the field and extract it
+                # with shift/mask. SMEM has no sub-dword loads, so whether a
+                # bare i16 load of the dispatch packet is widened to an
+                # s_load_dword depends on the surrounding use pattern (e.g. a
+                # zext to i64 -- Julia's default integer arithmetic --
+                # defeats it); the VMEM fallback is a per-wave read of
+                # uncached fine-grained queue memory that stalls every wave
+                # at kernel entry. A dword-aligned dword load is always
+                # SMEM-selectable.
+                aligned = offset & ~0x3
+                idx_ptr_i8 = inbounds_gep!(builder, T_int8, ptr, [ConstantInt(aligned)])
+                idx_ptr_i32 = bitcast!(builder, idx_ptr_i8, T_ptr_i32)
+                word = load!(builder, T_int32, idx_ptr_i32)
+                metadata(word)[LLVM.MD_invariant_load] = MDNode(LLVM.Metadata[])
+                idx = word
+                if (offset & 0x2) != 0
+                    idx = lshr!(builder, idx, ConstantInt(Int32(16)))
+                end
+                idx = and!(builder, idx, ConstantInt(Int32(0xffff)))
+                ret!(builder, idx)
+            else
+                idx_ptr_i8 = inbounds_gep!(builder, T_int8, ptr, [ConstantInt(offset)])
+                idx_ptr_T = bitcast!(builder, idx_ptr_i8, T_ptr_T)
+                idx_T = load!(builder, T_T, idx_ptr_T)
+                idx = zext!(builder, idx_T, T_int32)
 
-            # attach range metadata
-            metadata(idx_T)[LLVM.MD_range] = MDNode([
-                ConstantInt(T(range.start)),
-                ConstantInt(T(range.stop))])
-            ret!(builder, idx)
+                # attach range metadata
+                metadata(idx_T)[LLVM.MD_range] = MDNode([
+                    ConstantInt(T(range.start)),
+                    ConstantInt(T(range.stop))])
+                metadata(idx_T)[LLVM.MD_invariant_load] = MDNode(LLVM.Metadata[])
+                ret!(builder, idx)
+            end
         end
 
         call_function(llvm_f, UInt32)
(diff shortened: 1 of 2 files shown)


luraess:
Thanks - I'll look into it. Could you maybe report perf gains on a representative MWE (possibly in terms of effective memory throughput) before/after the fix and compare to what you are getting on CUDA hardware?

luraess:
I reproduced it on gfx1100 (RX 7900 XTX, ROCm 7.2.3, Julia 1.12.7) with help from robot and confirm the problem, and the fix addresses what's claimed. 

However, the code comment's explanation isn't quite right. It's worth correcting since it's what the next reader will rely on.

SMEM has no sub-dword loads and needs 4-byte alignment. `_dim` emits `load i16 ..., align 2` (LLVM.jl's default ABI alignment for `i16`). With a *single* dim read, LLVM's AMDGPU `widenLoad` combine widens it to a dword and selects `s_load_b32`. With **two or more**, the Load-Store Vectorizer merges the adjacent `i16` loads *first* into `load <3 x i16>, align 2` — and a 6-byte, 2-aligned load can't be SMEM, so it falls out as VMEM. This is confirmed with `-print-after=load-store-vectorizer`: identical IR apart from `align 2` (→ VMEM) vs `align 4` (→ SMEM).

So the trigger is the number of dim components read, not Int64 arithmetic. Int32 vs Int64 makes no difference, so the `zext`-to-i64 claim in the comment/commit message should be dropped.

The approach seems good. One smaller fix one could consider would just annotate the true alignment (`align 4` at offsets 4 and 8) which also fixes it in one line and keeps the `!range` metadata. But a kernel reading only `.y`+`.z` vectorizes into a load starting at offset 6, which is only 2-aligned, and stays VMEM. The shift/mask version covers that case too and doesn't depend on any LLVM heuristic.

**Test on my gfx1100:**

- Codegen: dispatch-packet VMEM loads gone in all of x / y / xy / xyz / xyz-Int32 / dims-only.
- `llc` on gfx90a, gfx942, gfx950, gfx1100 — same before/after behaviour on all four, so this isn't arch-specific.
- Runtime correctness: 3D index reconstruction, plus `workgroupDim`/`gridItemDim`/`gridGroupDim` across several groupsizes including 1024 are all correct.
- Static, 3D diffusion stencil: 166 → 163 instructions, two dispatch VMEM loads → one `s_load_b64`, `s_waitcnt` 12 → 10, same VGPR count and occupancy. The 16-bit field extraction moves from `v_and`/`v_lshrrev` (VALU) to `s_and`/`s_lshr` (SALU).
- Throughput, same stencil:

| | before | after | |
|---|---|---|---|
| n=128 | 25.7 µs / 653 GB/s | 23.4 µs / 718 GB/s | **+10%** |
| n=256 | 218.3 µs / 615 GB/s | 198.7 µs / 675 GB/s | **+10%** |
| n=512 | 1707.8 µs / 629 GB/s | 1704.2 µs / 630 GB/s | noise |
| 1D copy (control, reads only `.x`) | — | identical | — |


One caveat: the win appears for launch-overhead dominated kernel. Largest for short kernels and small grids, gone by n=512. So I'm not sure it alone explains an MI355X-vs-H200 gap on a large bandwidth-bound problem, and the before/after MWE numbers would still be interesting.

**Minor points:**

- The `UInt16` path drops the `!range` metadata. I measured no codegen regression from this (including `gridGroupDim`'s `udiv`), but it's a (deliberate) tradeoff and may need a line in the comment.
- `Int32(mask)` would throw for a 4-byte type. This is unreachable now since the branch is guarded on `aligned_sizeof(T) < 4`, but possibly brittle if a 32-bit field ever routes there.
- The test's `!occursin("global_load", gcn)` is fine for this kernel, but won't catch `buffer_load` on older targets. Possibly asserting on the absence of a dispatch-pointer-relative VMEM load specifically would be more robust.

sheevy:
ok, let me read through this.
My MWE shows much more severe performance diff: https://github.com/wave-photonics/amdgpu-workgroupdim-mwe

luraess:
Thanks for the link. I ran your MWE on my gfx1100.

Your ISA table reproduces exactly, via your own `packet_loads` probe:

```
dims read   unfixed                                  fixed
1 (x)       s_load_b32                               s_load_b32
2 (x,y)     global_load_b32                          s_load_b32
3 (x,y,z)   global_load_b32 + global_load_u16        s_load_b64
```

The throughput on my Radeon RDNA GPU differs from what you are getting on Instinct CDNA:

```
kernel        loads     unfixed    fixed   recovered
stream2/dim   2 -> 1      0.90x    0.88x       +1.4%
stream6/dim   2 -> 1      1.04x    1.05x       -1.1%
stream14/dim  2 -> 1      0.99x    1.00x       -1.0%
churn         2 -> 1      2.39x    1.68x      +42.3%
```

N=256, Float64, block 256. The streaming rows are a null here — `stream14/dim` is 0.99x against its own `arg` control *before* the fix, so on RDNA3 there is nothing there to recover. Your second axis, the one where more loads in flight means a larger cost, doesn't appear on this part at all.

The short-kernel row does reproduce, and solidly: 2.28-2.39x unfixed against 1.57-1.68x fixed over three runs each, or 8.6-9.2 → 6.4-6.5 µs per launch with the `arg` baseline steady at 3.6-4.1 µs. That lines up with the launch-overhead-shaped result I posted above, measured a completely different way.

2 items you may want to address:
- **`_dim` comment**: your MWE header already has the correct diagnosis, so the `zext`-to-i64 explanation in the in-code comment can just go.
- **Test**: port `dispatch_reg`/`packet_loads` from the MWE into `test/core/codegen.jl`. Reading the dispatch register off the kernel descriptor and matching source operands only is much better than grepping for `global_load`, and it survives targets where the fallback is `buffer_load`.

sheevy:
I observed this issue on mi355x (gfx950/cdna4), definitely not on strix halo (gfx1151/rdna3.5). I'm not sure if mi325x was problematic (gfx942/cdna3).
I will check you comments and improvement suggestions.

gbaraldi:
I think this is maybe due to the alignment being wrong

luraess:
This PR should be superseded by #1058

luraess:
@sheevy can you re test your slow MWE now that #1058 got merged from main branch of AMDGPU as this most likely supersedes this PR in a more generic way.

sheevy:
@luraess not at the moment, I'll get back to you on this next week

Issue #1055, opened 2026-08-26
@ROCStaticLocalArray's default zeroinit costs ~11x: every lane memsets the whole LDS array at kernel entry


gbaraldi:
`@ROCStaticLocalArray` (and `@ROCDynamicLocalArray`) default to `zeroinit=true`, and the zero-initialization is implemented in a way that costs an order of magnitude on any kernel that uses it.

**Measurement** (MI300A, ROCm 7.2.4, Julia 1.12.6): a 256-thread LDS tree reduction over 2²⁶ Float32, identical code except the macro argument:

| | time | effective bandwidth |
|---|---|---|
| `@ROCStaticLocalArray(Float32, 256)` (default) | 2,122.8 µs | 126 GB/s |
| `@ROCStaticLocalArray(Float32, 256, false)` | 187.8 µs | 1,430 GB/s |

**Why it is this expensive.** `zeroinit_lds!` (src/compiler/zeroinit_lds.jl) inserts a `memset` of the whole LDS global at kernel entry followed by `s_barrier`. Every work-item executes that memset, so for a 1 KiB array the ISA is 64 × `ds_write_b128` per wave, all 64 lanes writing the *same* address each time (`ds_write_b128 v1, v[2:5] offset:N` with `v1 = 0`), which serializes on bank conflicts — and that's repeated by every wave in the workgroup, on every launch. The prologue dwarfs the reduction itself.

**Related surfaces**
- `KernelAbstractions.@localmem` (`KA.SharedMemory` → `alloc_special` without zeroinit) and `src/kernels/mapreduce.jl` (`@ROCDynamicLocalArray(T, items, false)`) already avoid it, so library code is fine; only the user-facing macros pay by default.
- `@ROCDynamicLocalArray`'s zeroinit is a per-work-item serial loop over the whole array (`for idx in 1:prod(dims); DA[idx] = zero(T)`), same pathology.
- CUDA.jl's `CuStaticSharedArray` does not zero shared memory; neither does HIP's `__shared__`.

**Suggested fix**: default `zeroinit=false` (matching CUDA.jl/HIP semantics; document that LDS contents are undefined), and if zeroing is requested, distribute it — each work-item clears its strided slice (`for i in tid:groupsize:len`) before the barrier — instead of every lane clearing everything.

PR #1056, opened 2026-08-26
Run the AMDGPU attributor after optimization (kernel descriptors match hipcc)

gbaraldi:
While auditing AMDGPU.jl's codegen against hipcc output on an MI300A, the kernel descriptors stood out: a 1-D `vadd` compiled by AMDGPU.jl requests **every** implicit input — 8 user SGPRs (dispatch ptr, queue ptr, kernarg ptr, dispatch id), all three workgroup ids, and packed x/y/z workitem ids (`.amdhsa_system_vgpr_workitem_id 2`). hipcc emits 2 user SGPRs and `workitem_id 0` for the same kernel.

The difference is LLVM's `AMDGPUAttributor`: clang's pipeline runs it, GPUCompiler's never does. It proves which implicit inputs a kernel can never touch and records `amdgpu-no-*` attributes; the backend then stops requesting those inputs.

This PR runs it in `finish_ir!`, i.e. after optimization. Placement matters: in `finish_module!` (pre-optimization) it inferred nothing, because un-inlined helper calls block the interprocedural proof.

Effect on MI300A (gfx942, ROCm 7.2.4) for the probe kernels:

| kernel | before | after |
|---|---|---|
| 1-D vadd | queue_ptr 1, dispatch_id 1, workitem_id 2, 15 SGPRs | queue_ptr 0, dispatch_id 0, workitem_id 0, 12 SGPRs |
| LDS reduction (no `workgroupDim`) | dispatch_ptr 1, ... | also `amdgpu-no-dispatch-ptr` |
| 3-D kernel | workitem_id 2 | workitem_id 2 (correctly kept) |

Validated on hardware that kernels which *do* use implicit arguments keep working: `@rocprintf` (hostcall — queue ptr + dispatch id), a device-side bounds exception surfacing on the host, 3-D indexing, LDS + `sync_workgroup`, and the broadcast/mapreduce/rocBLAS library paths.

Kernel descriptors before/after and the probe script are in the issue trail of #1055 / this branch's commit message.

### DIFF
--- a/src/compiler/codegen.jl
+++ b/src/compiler/codegen.jl
@@ -408,15 +408,8 @@ function run_and_collect(cmd)
     return proc, log
 end
 
-# Infer implicit-argument usage after optimization. LLVM's AMDGPUAttributor
-# proves which implicit kernel inputs (dispatch/queue pointer, dispatch id,
-# workgroup/workitem ids for y/z, hostcall/heap pointers, ...) a kernel can
-# never touch and records that as `amdgpu-no-*` attributes; the backend then
-# stops requesting those SGPR/VGPR inputs in the kernel descriptor. Clang runs
-# this in its default pipeline; GPUCompiler's pipeline never does, so every
-# kernel was compiled as if it used all of them (8 user SGPRs, all workgroup
-# ids, packed workitem ids for x/y/z). It has to run after optimization —
-# before inlining, un-inlined helper calls block the interprocedural proof.
+# Run amdgpu-attributor so that the amdgpu specific attributes are added
+# This reduces some register usage
 function GPUCompiler.finish_ir!(
     @nospecialize(job::HIPCompilerJob), mod::LLVM.Module, entry::LLVM.Function,
 )
--- a/src/compiler/codegen.jl
+++ b/src/compiler/codegen.jl
@@ -407,3 +407,29 @@ function run_and_collect(cmd)
     log = strip(fetch(reader))
     return proc, log
 end
+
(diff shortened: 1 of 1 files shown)


src/compiler/codegen.jl
```diff
@@ -394,3 +394,37 @@ function run_and_collect(cmd)
...
+    # optimization, so skip it where the parser doesn't know it rather than error.
+    try
+        @dispose pb=NewPMPassBuilder() begin
+            add!(pb, "amdgpu-attributor")
+            run!(pb, mod, tm)
+        end
+    catch err
+        err isa LLVM.LLVMException || rethrow()
+        @debug "amdgpu-attributor pass unavailable on LLVM $(LLVM.version()); skipping"
+    end
```
simeonschaub: Doing this with a try-catch seems a little ugly, should we add a check based on `Base.libllvm_version` instead?

simeonschaub (commented):
I'm wondering whether this should go into GPUCompiler instead, but I don't really have a strong preference either way. Overall SGTM

simeonschaub (approved):

luraess:
The remaining failure relates to the Int128 miscompilation https://buildkite.com/julialang/amdgpu-dot-jl/builds/4098#01a048dc-daaf-440e-adb2-2d8c8cf14060/L1372

simeonschaub:
Might be https://github.com/llvm/llvm-project/pull/213523 again

luraess:
@gbaraldi Can you rebase on main as now the failing Int128 test should be excluded

luraess:
Thanks!

PR #1057, opened 2026-08-26
Fix off-by-one in indexing !range metadata

luraess:
LLVM reads `!range` as the half-open `[lo, hi)`, but `_index` and `_dim` pass inclusive Julia ranges straight through. The topmost legal value falls outside the declared range, so comparisons against it fold to false.

On `main` at `groupsize=1024`:

```julia
i = workitemIdx().x
(i == 1024) && (X[1] = 1)
(i >= 1024) && (X[2] = 1)
(workgroupDim().x == 1024) && (X[3] = 1)
(gridItemDim().x == 1024) && (X[4] = 1)
# Int32[0, 0, 0, 1]  -- expected Int32[1, 1, 1, 1]
```

Lane 1024 does run — stores aren't dropped, so `fill!`/broadcast were never corrupting data; only the comparisons vanish.

`_dim` has a second bug: it reads *sizes*, but every call site hands it an *index* range, so `workgroupDim` declared `[0, 1023)` where the truth is `[1, 1025)`.

## Changes

Both go through `_range_metadata(T, range)`, which adds one **in `T`** — a full-width range then wraps to `[lo, 0)`, which LLVM accepts — and returns `nothing` when `lo == hi`, which LLVM rejects. The `_dim` call sites pass size ranges.

| intrinsic | before | after |
|---|---|---|
| `workitemIdx` | `[0, 1023)` | `[0, 1024)` |
| `workgroupIdx` | `[0, 4294967294)` | `[0, 0xffffffff)` |
| `workgroupDim` | `[0, 1023)` | `[1, 1025)` |
| `gridItemDim` | `[0, 4294967294)` | `[1, 0)` wrapping (nonzero) |

## Testing

`core` + `device` pass on gfx1100 (ROCm 7.2.3, Julia 1.12.7). The new testset fails on `main` with `Int32[0, 0, 0, 1]`, so it guards the bug rather than just describing it. The wrapping `[1, 0)` form is checked against `opt`: the verifier accepts it, `x == 0` folds, `x == typemax` stays reachable.

### DIFF
--- a/src/device/gcn/indexing.jl
+++ b/src/device/gcn/indexing.jl
@@ -1,5 +1,13 @@
 # Indexing and dimensions
 
+# LLVM's !range is half-open [lo, hi); the ranges below are inclusive.
+function _range_metadata(::Type{T}, range) where T
+    lo = T(range.start)
+    hi = T(range.stop) + one(T)
+    lo == hi && return nothing
+    return MDNode([ConstantInt(lo), ConstantInt(hi)])
+end
+
 @device_function @generated function _index(::Val{fname}, ::Val{name}, ::Val{range}) where {fname, name, range}
     @dispose ctx=Context() begin
         T_int32 = LLVM.Int32Type()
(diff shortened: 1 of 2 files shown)


luraess:
cscs-ci run

luraess:
cscs-ci run

luraess:
Is this good to go @gbaraldi or superseded by one of your incoming PRs?

PR #1058, opened 2026-08-26
Read workgroup/grid dimensions from hidden kernargs instead of the dispatch packet (2x on short kernels)

gbaraldi:
`workgroupDim`/`gridItemDim` (and everything built on them — the canonical global-index computation) read the AQL dispatch packet through `llvm.amdgcn.dispatch.ptr`. That packet lives in host-coherent queue memory, so **every wave of every kernel pays an uncached SMEM round trip in its first few instructions**. On short kernels this dominates.

Measured on MI300A (gfx942, ROCm 7.2.4) with the rocm-examples Floyd–Warshall kernel (2048 nodes → 2048 dependent launches of a 39-instruction kernel):

| variant | per launch |
|---|---|
| `workgroupDim()` from the dispatch packet (before) | 21.5 µs |
| same kernel, block dims hard-coded (no packet read) | 11.2 µs |
| hipcc's kernel, launched from the same Julia harness | 10.9 µs |
| **`workgroupDim()` from hidden kernargs (this PR)** | **11.4 µs** |

The ISA of the Julia and hipcc kernels is otherwise identical (39 instructions each, same register counts); the only difference was `s_load_dword s10, s[0:1], 0x4` off the dispatch pointer vs clang's `s_load_dword s10, s[0:1], 0x24` off the kernarg segment — `hidden_group_size_x` at explicit-args + 12.

This PR does what clang does: dimensions come from the **hidden kernel arguments** (code object v5+, which GPUCompiler already emits — v6 here), which the runtime writes into the kernarg segment in device memory: `hidden_block_count_{x,y,z}` (u32 at 0/4/8) and `hidden_group_size_{x,y,z}` (u16 at 12/14/16) behind `llvm.amdgcn.implicitarg.ptr`. Consequences:

- `gridGroupDim` is a direct load instead of `udiv(grid_size, group_size)`; `gridItemDim` is `block_count * group_size` (launches are whole workgroups, so this equals the packet's `grid_size`).
- Kernels get `"amdgpu-implicitarg-num-bytes"="256"` so the backend emits the hidden-arg metadata the runtime keys on (`kernarg_segment_size` grows by 256 bytes; verified `hidden_block_count_x`/`hidden_group_size_x` appear and the runtime fills them).
- The loads carry their real alignment, `!range` and `!invariant.load`, so they stay SMEM-selectable — this supersedes the concern in #1054 for the common case (no dispatch-packet loads remain to widen).
- With #1056 (attributor) on top, the dispatch pointer user SGPR disappears entirely since nothing references `dispatch.ptr` anymore.

Broader effect on MI300A: BabelStream Copy 3,306 → 3,601 GB/s (95% of hipcc), Mul 3,377 → 3,602 (97%); rocm-examples bitonic sort 18.4 → 16.6 ms (hipcc 16.0), 5×5 convolution 123 → 110 µs; LDS tiled matmul +2.5%. Correctness validated on the full rocm-examples Applications port (histogram, prefix-sum, convolution, bitonic sort, Floyd–Warshall — all checked against references).

---

**Second commit — bound workgroup indices to `Int32` so checked conversions fold.**

The `!range` on `workgroupIdx` (and now `gridGroupDim`) spanned the full `UInt32`, so the natural `Int32(workgroupIdx().x)` kept its `InexactError` check — and since `throw_inexacterror` is a real call, any kernel indexing that way grew a call frame and SGPR save/restore (the Floyd–Warshall port above: 281 instructions vs 39, 2.8× slower). `workitemIdx`/`workgroupDim` never had this problem: their ranges fit `Int32` and LLVM folds the check.

HSA's `grid_size` fields are `UInt32` work-items, so a workgroup index can only exceed `typemax(Int32)` for more than 2³¹ *single-work-item* groups. `Runtime.launch` now rejects that, and the tighter bound is recorded as `!range`. With the **unchanged** checked idiom, MI300A: Floyd–Warshall 28.9 → 11.4 µs/launch (hipcc 10.9), zero throw calls, 33 instructions; bitonic sort 18.4 → 16.7 ms (hipcc 15.7). Return type stays `UInt32`.

### DIFF
--- a/src/compiler/codegen.jl
+++ b/src/compiler/codegen.jl
@@ -103,10 +103,16 @@ function GPUCompiler.finish_module!(
 
         # TODO add convergent, mustprogress, willreturn attributes?
 
+        # request the full hidden-argument block (code object v5+): workgroup and
+        # grid dimensions are read from it (see device/gcn/indexing.jl), and the
+        # metadata this emits is what makes the runtime fill it in at launch
+        implicitarg_attr = StringAttribute("amdgpu-implicitarg-num-bytes", "256")
+
         attrs = LLVM.function_attributes(entry)
         push!(attrs, target_cpu_attr)
         push!(attrs, target_features_attr)
         push!(attrs, atomic_attr)
+        push!(attrs, implicitarg_attr)
     end
 
     # Workaround for the lack of zeroinitializer support for LDS.
--- a/src/compiler/codegen.jl
+++ b/src/compiler/codegen.jl
@@ -104,8 +104,7 @@ function GPUCompiler.finish_module!(
         # TODO add convergent, mustprogress, willreturn attributes?
 
         # request the full hidden-argument block (code object v5+): workgroup and
-        # grid dimensions are read from it (see device/gcn/indexing.jl), and the
-        # metadata this emits is what makes the runtime fill it in at launch
+        # grid dimensions are read from it (see device/gcn/indexing.jl),
         implicitarg_attr = StringAttribute("amdgpu-implicitarg-num-bytes", "256")
 
         attrs = LLVM.function_attributes(entry)
--- a/src/device/gcn/helpers.jl
+++ b/src/device/gcn/helpers.jl
@@ -1,7 +1,3 @@
-# HSA dispatch packet offsets
-_packet_names = fieldnames(HSA.KernelDispatchPacket)
-_packet_offsets = fieldoffset.(HSA.KernelDispatchPacket, 1:length(_packet_names))
-
 # which Julia types map to a given ROC lib type
 const fntypes = Dict{Type,Symbol}(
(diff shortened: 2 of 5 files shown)


luraess (changes requested):
LGTM - thanks! Could you unfable the inline comments keeping it to the min and not case/issue specific - thanks.

luraess:
Some conflicts need resolving before this can be finalised.

luraess:
Is #1056 needed as companion for this PR @gbaraldi ?

luraess:
@gbaraldi ping on that :-)

gbaraldi:
No

luraess:
> No

Thanks. I will thus merge this one.

luraess:
Thanks!

PR #1063, opened 2026-09-03
Claim the bundle's vendor libraries, and drop the rocSPARSE version probe

luraess:
EDIT: rolled-back most of #1001 

In artifact mode on a host with another ROCm on `LD_LIBRARY_PATH` (here a CSCS uenv), `versioninfo()` reported rocSPARSE as `err` and the core test asserting the probe returns a version failed. The cause is not rocSPARSE.

`PRELOAD_LIBRARIES` claimed the runtime closure but no vendor library, and AMDGPU claims those lazily on the first `ccall`, so whichever is touched first wins. TheRock's libraries carry `DT_RUNPATH`, which `LD_LIBRARY_PATH` outranks, so touching rocSPARSE first pulled its dependencies from the environment. Measured in artifact mode after `rocSPARSE.version()`:

```
rocsparse -> <artifact>/lib/librocsparse.so
rocblas   -> /user-environment/env/default/lib/librocblas.so.5
roctx     -> /user-environment/env/default/lib/libroctx64.so.4
foreign_libraries() = String[]
```

`FOREIGN_LIBRARY_NAMES` listed only runtime libraries, so `versioninfo()` printed "Foreign ROCm libraries loaded: none" while two rocBLAS versions were live in the process.

The isolated probe from #1051 hit the same thing: a bare child has claimed nothing at all, so it resolved the bundle's `librocsparse` against uenv HIP 7.2.3 and `rocsparse_create_handle` segfaulted, which is what surfaced as `err`.

### Changes

- `PRELOAD_LIBRARIES` also claims `libamdhip64`, `libroctx64`, `libhipblaslt` and `librocblas` — the vendor libraries that other vendor libraries pull in (`librocsparse` needs `librocblas` and `libroctx64`; `librocblas` needs `libhipblaslt`). Order is load-bearing: HSA must precede HIP, or HIP's `libhsa-runtime64.so.1` binds outside the bundle and the dlopen fails with `undefined symbol: hsa_amd_queue_create, version ROCR_1`.
- `FOREIGN_LIBRARY_NAMES` gains the vendor libraries, so a mixed process is reported instead of passing as clean.
- The out-of-process version probe is removed (`_version_subprocess`, `_rocsparse_version_isolated`, the cache global and its `__init__` reset). With the libraries claimed correctly there is nothing left for it to protect against, and it was a lot of machinery for one cell of a table.
- `versioninfo` queries rocSPARSE in-process like every other library. The shared `_ver` helper now guards *all* of them, so a failing query shows `err` instead of taking down the whole report — previously none were guarded.

### Result

On beverin (2x MI300A, artifact ROCm under uenv), before -> after:

| | before | after |
|---|---|---|
| `versioninfo` rocSPARSE | `err` + misleading warning | `5.0.0` |
| `librocblas` / `libroctx64` / `libhipblaslt` | uenv | bundle |
| `foreign_libraries()` | `String[]` (wrong) | `String[]` (correct) |
| preloaded libraries | 41 | 45 |

### Trade-off

A SIGSEGV inside `rocsparse_create_handle` (#920) again takes down the session rather than degrading to `err`. Two things make that acceptable: the preload fix removes the mixed-library cause that produced the crash here, and `versioninfo` already initialises HIP and enumerates devices in-process, so a genuinely broken install fails before the rocSPARSE row either way. `try`/`catch` still covers every non-fatal failure.

### DIFF
--- a/ROCm_Runtime/src/ROCm_Runtime.jl
+++ b/ROCm_Runtime/src/ROCm_Runtime.jl
@@ -67,6 +67,8 @@ end
 
 # Claimed from the bundle before LD_LIBRARY_PATH can resolve them elsewhere.
 # Leaves first, as (subdirectory of the library directory, prefix, open all).
+# HSA must precede HIP, or HIP's `libhsa-runtime64.so.1` binds outside the
+# bundle; the vendor libraries here are ones other vendor libraries pull in.
 const PRELOAD_LIBRARIES = [
     ("rocm_sysdeps/lib", "librocm_sysdeps_",        true),
     ("llvm/lib",         "libLLVM",                 false),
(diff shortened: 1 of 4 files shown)


simeonschaub:
I'm really not a big fan of introducing even more complexity just for the librocsparse version check. With #1028, we might want to revisit whether the check is truly needed anymore, since now the user would have to explicitly request AMDGPU.jl to use a broken system ROCm install for this to matter. I personally think it's not too unexpected that Julia would crash in such a situation, so I'd propose reverting #1001 altogether as part of #1028

luraess:
Yeah, that's the other alternative, to remove all this machinery with the one caveat is that we may get inaccurate version for rocSPARSE if we ever encounter using it from system. I can do it here.

luraess:
This is now mostly rolling back #1001 but adding some guards.

PR #1065, opened 2026-09-03
Handle hipTENSOR 2.4: fix the version decode, reject the layouts it miscomputes

luraess:
hipTENSOR 2.4 (the version in the ROCm 10.0 bundles) no longer uses the mode labels to line a contraction's operands up. It reads every descriptor positionally, and computes the right result only when the operands are packed column-major and indexed as

* `A = [M…, K…]` — A's free modes, then the contracted ones
* `B = [N…, K…]`
* `C = D = [M…, N…]` — A's free modes, then B's

Every other layout is **silently miscomputed**: `hiptensorContract` returns `HIPTENSOR_STATUS_SUCCESS` and writes a plausible but wrong result. hipTENSOR 2.2 handled arbitrary mode orders, which is what the wrapper was written against.

Of the 120 shapes `test/hiptensor/contractions.jl` generates, 115 came out wrong on an MI300A, with relative errors of 0.3–1.2. Reordering the descriptors alone fixes only 80 of them; all 120 are correct once every operand is packed and canonical.

Two related changes in 2.4:

* `HIPTENSOR_OP_CONJ` on a contraction's input is now rejected — `hiptensorCreatePlan` fails with `HIPTENSOR_STATUS_EXECUTION_FAILED`. On 2.2 it was accepted and silently ignored.
* `hiptensorGetVersion()` returns `2004000`, i.e. `major * 1_000_000 + minor * 1_000 + patch`. `version()` decoded it cuTENSOR-style and reported the library as `v"200.40.0"`.

## Changes

* fix the `version()` decode, so hipTENSOR 2.4.0 reports `v"2.4.0"`
* add `check_contraction_layout`, called from `plan_contraction`: on hipTENSOR ≥ 2.4, reject any layout the library would miscompute, naming the index order each operand needs. Below 2.4 it returns immediately, so 2.2 is untouched.
* document both restrictions in `docs/src/libraries/tensor.md`
* tests use random mode orders below 2.4 and the canonical layout at or above it, expect `hipTENSORError` from `OP_CONJ` on 2.4, and gain a test asserting the guard fires

Making arbitrary layouts work on 2.4 would mean materialising permuted copies of every operand — a copy of `A`, `B` and `C` on most calls — so that is left until upstream says whether the layout requirement is intended.

## Testing

Verified on an MI300A (gfx942) against hipTENSOR 2.4.0 from the ROCm 10.0 `gfx94X-dcgpu` bundle: `test/hiptensor/contractions.jl` passes 433/433, no failures or broken tests.

Not tested against hipTENSOR 2.2. That path is unchanged by construction.

### DIFF
--- a/docs/src/libraries/tensor.md
+++ b/docs/src/libraries/tensor.md
@@ -8,10 +8,7 @@ which can be used to perform operations on high-dimensional arrays. However, the
   operator passed as `opABC` to `A` and `B`, and the one passed as `opAB` to that intermediate result and `C`. This is currently handled by our wrapper.
 - Unlike NVIDIA's `cuTENSOR`, `hipTENSOR` wants a *complex* compute descriptor for complex operands: pairing e.g. `ComplexF32` tensors with `HIPTENSOR_COMPUTE_DESC_32F` makes
   `hiptensorCreatePlan` fail with `HIPTENSOR_STATUS_EXECUTION_FAILED`
-- from `hipTENSOR` 2.4 on (the version in the ROCm 10.0 bundles, for instance), the contraction kernels no longer use the mode labels to work out how their operands line up:
-  they read every descriptor positionally, and only compute the right result when `A` is indexed by its free modes followed by the contracted ones, `B` likewise, and `C` and `D`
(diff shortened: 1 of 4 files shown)


luraess:
Changes were comments only. Merging into main in order to have it avail for #1028

luraess:
cscs-ci run

PR #1066, opened 2026-09-04
Skip Int128 in the GPUArrays test suite on LLVM 18

maleadt:
GPUArrays.jl#723 adds Int128 to the test suite's default element types on Julia 1.12+, but Julia 1.12's in-tree LLVM 18 miscompiles Int128 arithmetic on AMDGPU (#1002). Add an explicit `supported_eltypes` overload for ROCArray, like oneAPI.jl does, and only include Int128 on LLVM 19 or newer.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -80,6 +80,19 @@ init_code = quote
     include($gpuarrays_testsuite)
     testf(f, xs...; kwargs...) = TestSuite.compare(f, AMDGPU.ROCArray, xs...; kwargs...)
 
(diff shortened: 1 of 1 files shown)


simeonschaub:
Note that 1.13 and nightly still fail. We are using AMDGPU_LLVM_Backend_jll 22.1.8 regardless of Julia's LLVM version already, we'll need something like https://github.com/JuliaPackaging/Yggdrasil/pull/14334 to fix the actual issues. For now, it should be fine to disable Int128 tests regardless of Julia version

PR #1067, opened 2026-09-04
Exclude Int128 from testing on linalg/core

luraess:
Possibly workaround #1066 according to https://github.com/JuliaGPU/AMDGPU.jl/pull/1066#issuecomment-5538187568

### DIFF
--- a/test/core/rocarray_broadcast.jl
+++ b/test/core/rocarray_broadcast.jl
@@ -59,3 +59,18 @@ end
     @test Array(test_kernel.(ROCArray(M), Int128(10))) == test_kernel.(M, Int128(10))
     AMDGPU.synchronize()
 end
+
+# https://github.com/JuliaGPU/AMDGPU.jl/issues/1002
+# The multiply-add that GPUArrays' `axpy!`/`axpby!` lower to is still miscompiled on
+# LLVM 20, which is why `runtests.jl` drops Int128 from the `gpuarrays/linalg/core`
+# element types.
+if Base.libllvm_version >= v"19"
+    @testset "Int128 axpby! miscompilation" begin
+        a, b = rand(Int128), rand(Int128)
+        x, y = rand(Int128, 5), rand(Int128, 5)
+        gx, gy = ROCArray(x), ROCArray(y)
+        gy .= gx .* a .+ gy .* b
+        @test_broken Array(gy) == x .* a .+ y .* b
+        AMDGPU.synchronize()
+    end
+end
--- a/test/core/rocarray_broadcast.jl
+++ b/test/core/rocarray_broadcast.jl
@@ -61,11 +61,14 @@ end
 end
 
 # https://github.com/JuliaGPU/AMDGPU.jl/issues/1002
-@testset "Int128 axpby! miscompilation" begin
-    a, b = rand(Int128), rand(Int128)
-    x, y = rand(Int128, 5), rand(Int128, 5)
-    gx, gy = ROCArray(x), ROCArray(y)
-    gy .= gx .* a .+ gy .* b
-    @test_broken Array(gy) == x .* a .+ y .* b
-    AMDGPU.synchronize()
+# Gated on the same condition that puts Int128 into the `linalg/core` eltypes:
+if int128_supported
+    @testset "Int128 axpby! miscompilation" begin
+        a, b = rand(Int128), rand(Int128)
+        x, y = rand(Int128, 5), rand(Int128, 5)
+        gx, gy = ROCArray(x), ROCArray(y)
+        gy .= gx .* a .+ gy .* b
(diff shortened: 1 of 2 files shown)


test/core/rocarray_broadcast.jl
```diff
@@ -59,3 +59,18 @@ end
     @test Array(test_kernel.(ROCArray(M), Int128(10))) == test_kernel.(M, Int128(10))
     AMDGPU.synchronize()
 end
+
+# https://github.com/JuliaGPU/AMDGPU.jl/issues/1002
+# The multiply-add that GPUArrays' `axpy!`/`axpby!` lower to is still miscompiled on
+# LLVM 20, which is why `runtests.jl` drops Int128 from the `gpuarrays/linalg/core`
+# element types.
```
simeonschaub: This comment doesn't really make sense, since we're using the external LLVM backend now, which is always LLVM 22. I would just delete it

test/core/rocarray_broadcast.jl
```diff
@@ -59,3 +59,18 @@ end
...
+if Base.libllvm_version >= v"19"
```
simeonschaub: Why this check of the LLVM version? It's broken everywhere atm, so I don't see a reason to restrict this to only LLVM >= 19

test/runtests.jl
```diff
@@ -93,6 +93,13 @@ init_code = quote
     end
     TestSuite.supported_eltypes(::Type{<:AMDGPU.ROCArray}) = eltypes
 
+    # LLVM 20 still miscompiles the Int128 multiply-add that GPUArrays' `axpy!` and
+    # `axpby!` lower to, so keep Int128 out of that suite. `core/rocarray_broadcast`
+    # carries a `@test_broken` that starts failing once the miscompilation is fixed.
```
simeonschaub: Again, this comment is wrong. Maybe just referring to #1002 is enough?

simeonschaub (changes requested):

test/core/rocarray_broadcast.jl
luraess: I thought the Yggdrasil PR fixed it for `< v"19"`. But can remove the check

luraess: Seems to be the case @simeonschaub given the test d actually not fail on Julia 1.10 and 1.11 (e.g. https://buildkite.com/julialang/amdgpu-dot-jl/builds/4144#01a07d8e-68c7-4453-adbf-707fcffb6239/L1248)

test/runtests.jl
simeonschaub:
```suggestion
    const eltypes = [Int16, Int32, Int64, # TODO: add Int128 once #1002 is addressed
```

test/runtests.jl
```diff
@@ -93,6 +93,12 @@ init_code = quote
     end
     TestSuite.supported_eltypes(::Type{<:AMDGPU.ROCArray}) = eltypes
 
+    # Miscompilation of Int128 arithmetic on AMDGPU (JuliaGPU/AMDGPU.jl#1002).
+    # `core/rocarray_broadcast` carries a `@test_broken` to track this.
+    TestSuite.supported_eltypes(
+        ::Type{<:AMDGPU.ROCArray}, ::typeof(TestSuite.test_linalg_core),
+    ) = filter(!=(Int128), eltypes)
```
simeonschaub:
```suggestion
```

test/core/rocarray_broadcast.jl
```diff
@@ -59,3 +59,16 @@ end
...
+if int128_supported
+    @testset "Int128 axpby! miscompilation" begin
+        a, b = rand(Int128), rand(Int128)
+        x, y = rand(Int128, 5), rand(Int128, 5)
+        gx, gy = ROCArray(x), ROCArray(y)
+        gy .= gx .* a .+ gy .* b
+        @test_broken Array(gy) == x .* a .+ y .* b
+        AMDGPU.synchronize()
+    end
+end
```
simeonschaub:
```suggestion
@testset "Int128 axpby! miscompilation" begin
    a, b = rand(Int128), rand(Int128)
    x, y = rand(Int128, 5), rand(Int128, 5)
    gx, gy = ROCArray(x), ROCArray(y)
    gy .= gx .* a .+ gy .* b
    @test broken = Base.libllvm_version >= v"18" Array(gy) == x .* a .+ y .* b
    AMDGPU.synchronize()
end
```

simeonschaub (changes requested):

test/runtests.jl
simeonschaub: Hmm, looks like it did not apply correctly. I actually meant to just delete lines 87 to 100

simeonschaub (approved):
Thanks!

PR #1069, opened 2026-09-08
Guard empty-array launch in out-of-place reverse

### LINKED ISSUE #1068: reverse of an empty ROCArray throws hipErrorInvalidValue
Issue: `reverse` on an empty `ROCArray` (and the `dims` form) throws `HIPError(code hipErrorInvalidValue)` instead of returning an empty array the way `Base.reverse` does. The out-of-place `_reverse!` launches with `gridsize = cld(length(x), groupsize) = cld(0, 256) = 0`, and `hipModuleLaunchKernel` rejects the zero grid (gridDimX = 0). The in-place `reverse!` is spared by its early return on empty input.

MWE:
```julia
using AMDGPU

reverse(ROCArray(Float32[]))                             # ERROR: hipErrorInvalidValue
reverse(ROCArray(Matrix{Float32}(undef, 0, 3)); dims=1)  # same
reverse(ROCArray(Float32[1, 2, 3]))                      # OK, returns [3, 2, 1]
```

Reproduces on AMDGPU 2.3.0 through 2.8.0 (gfx1200, ROCm 7.2), for every eltype tested (Float32, Float64, ComplexF32, Int32) and for empty matrices with any `dims`.

shreyas-omkar:
Fixes #1068

### DIFF
--- a/src/kernels/reverse.jl
+++ b/src/kernels/reverse.jl
@@ -83,6 +83,7 @@ function _reverse!(
 
     groupsize = 256
     gridsize = cld(length(x), groupsize)
+    iszero(gridsize) && return
     @roc groupsize=groupsize gridsize=gridsize _kernel!(y, x)
 end
 
@@ -118,5 +119,6 @@ function _reverse!(x::AnyROCArray{T, N}; dims=1:ndims(x)) where {T, N}
 
     groupsize = 256
     gridsize = cld(prod(reduced_sz), groupsize)
+    iszero(gridsize) && return
     @roc groupsize=groupsize gridsize=gridsize _kernel!(x)
 end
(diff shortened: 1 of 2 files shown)


luraess:
Thanks. Maybe one could add one test as well?

shreyas-omkar:
Good idea, done - added an `empty` testset in `test/hip_rocarray/reverse.jl` covering empty vector and empty matrix (0xN, Nx0) reverse for `dims` `:`, 1, 2, both out-of-place and in-place. Verified it errors on the pre-fix version and passes with the guard (116/116 in the reverse suite locally on gfx1200).

luraess:
Thanks!

PR #1070, opened 2026-09-08
plan handle cache fix 

### LINKED ISSUE #1053: `rocFFT`: plan handle cache (`IDLE_HANDLES`) never evicts distinct-shape plans, leaking GPU memory
## Questionnaire

1. Does ROCm work for you outside of Julia, e.g. C/C++/Python?
 yes

2. Post output of `rocminfo`. (Agent 5 is about the GPU)
```
ROCk module version 6.10.5 is loaded
=====================    
HSA System Attributes    
=====================    
Runtime Version:         1.14
Runtime Ext Version:     1.6
System Timestamp Freq.:  1000.000000MHz
Sig. Max Wait Duration:  18446744073709551615 (0xFFFFFFFFFFFFFFFF) (timestamp count)
Machine Model:           LARGE                              
System Endianness:       LITTLE                             
Mwaitx:                  DISABLED
DMAbuf Support:          YES

==========               
HSA Agents               
==========               
*******                  
Agent 1                  
*******                  
  Name:                    AMD EPYC 7A53 64-Core Processor    
  Uuid:                    CPU-XX                             
  Marketing Name:          AMD EPYC 7A53 64-Core Processor    
  Vendor Name:             CPU                                
  Feature:                 None specified                     
  Profile:                 FULL_PROFILE                       
  Float Round Mode:        NEAR                               
  Max Queue Number:        0(0x0)                             
  Queue Min Size:          0(0x0)                             
  Queue Max Size:          0(0x0)                             
  Queue Type:              MULTI                              
  Node:                    0                                  
  Device Type:             CPU                                
  Cache Info:              
    L1:                      32768(0x8000) KB                   
  Chip ID:                 0(0x0)                             
  ASIC Revision:           0(0x0)                             
  Cacheline Size:          64(0x40)                           
  Max Clock Freq. (MHz):   2000                               
  BDFID:                   0                                  
  Internal Node ID:        0                                  
  Compute Unit:            32                                 
  SIMDs per CU:            0                                  
  Shader Engines:          0                                  
  Shader Arrs. per Eng.:   0                                  
  WatchPts on Addr. Ranges:1                                  
  Memory Properties:       
  Features:                None
  Pool Info:               
    Pool 1                   
      Segment:                 GLOBAL; FLAGS: FINE GRAINED        
      Size:                    131302256(0x7d38370) KB            
      Allocatable:             TRUE                               
      Alloc Granule:           4KB                                
      Alloc Recommended Granule:4KB                                
      Alloc Alignment:         4KB                                
      Accessible by all:       TRUE                               
    Pool 2                   
      Segment:                 GLOBAL; FLAGS: EXTENDED FINE GRAINED
      Size:                    131302256(0x7d38370) KB            
      Allocatable:             TRUE                               
      Alloc Granule:           4KB                                
      Alloc Recommended Granule:4KB                                
      Alloc Alignment:         4KB                                
      Accessible by all:       TRUE                               
    Pool 3                   
      Segment:                 GLOBAL; FLAGS: KERNARG, FINE GRAINED
      Size:                    131302256(0x7d38370) KB            
      Allocatable:             TRUE                               
      Alloc Granule:           4KB                                
      Alloc Recommended Granule:4KB                                
      Alloc Alignment:         4KB                                
      Accessible by all:       TRUE                               
    Pool 4                   
      Segment:                 GLOBAL; FLAGS: COARSE GRAINED      
      Size:                    131302256(0x7d38370) KB            
      Allocatable:             TRUE                               
      Alloc Granule:           4KB                                
      Alloc Recommended Granule:4KB                                
      Alloc Alignment:         4KB                                
      Accessible by all:       TRUE                               
  ISA Info:                
*******                  
Agent 2                  
*******                  
  Name:                    AMD EPYC 7A53 64-Core Processor    
  Uuid:                    CPU-XX                             
  Marketing Name:          AMD EPYC 7A53 64-Core Processor    
  Vendor Name:             CPU                                
  Feature:                 None specified                     
  Profile:                 FULL_PROFILE                       
  Float Round Mode:        NEAR                               
  Max Queue Number:        0(0x0)                             
  Queue Min Size:          0(0x0)                             
  Queue Max Size:          0(0x0)                             
  Queue Type:              MULTI                              
  Node:                    1                                  
  Device Type:             CPU                                
  Cache Info:              
    L1:                      32768(0x8000) KB                   
  Chip ID:                 0(0x0)                             
  ASIC Revision:           0(0x0)                             
  Cacheline Size:          64(0x40)                           
  Max Clock Freq. (MHz):   2000                               
  BDFID:                   0                                  
  Internal Node ID:        1                                  
  Compute Unit:            32                                 
  SIMDs per CU:            0                                  
  Shader Engines:          0                                  
  Shader Arrs. per Eng.:   0                                  
  WatchPts on Addr. Ranges:1                                  
  Memory Properties:       
  Features:                None
  Pool Info:               
    Pool 1                   
      Segment:                 GLOBAL; FLAGS: FINE GRAINED        
      Size:                    132110904(0x7dfda38) KB            
      Allocatable:             TRUE                               
      Alloc Granule:           4KB                                
      Alloc Recommended Granule:4KB                                
      Alloc Alignment:         4KB                                
      Accessible by all:       TRUE                               
    Pool 2                   
      Segment:                 GLOBAL; FLAGS: EXTENDED FINE GRAINED
      Size:                    132110904(0x7dfda38) KB            
      Allocatable:             TRUE                               
      Alloc Granule:           4KB                                
      Alloc Recommended Granule:4KB                                
      Alloc Alignment:         4KB                                
      Accessible by all:       TRUE                               
    Pool 3                   
      Segment:                 GLOBAL; FLAGS: KERNARG, FINE GRAINED
      Size:                    132110904(0x7dfda38) KB            
      Allocatable:             TRUE                               
      Alloc Granule:           4KB                                
      Alloc Recommended Granule:4KB                                
      Alloc Alignment:         4KB                                
      Accessible by all:       TRUE                               
    Pool 4                   
      Segment:                 GLOBAL; FLAGS: COARSE GRAINED      
      Size:                    132110904(0x7dfda38) KB            
      Allocatable:             TRUE                               
      Alloc Granule:           4KB                                
      Alloc Recommended Granule:4KB                                
      Alloc Alignment:         4KB                                
      Accessible by all:       TRUE                               
  ISA Info:                
*******                  
Agent 3                  
*******                  
  Name:                    AMD EPYC 7A53 64-Core Processor    
  Uuid:                    CPU-XX                             
  Marketing Name:          AMD EPYC 7A53 64-Core Processor    
  Vendor Name:             CPU                                
  Feature:                 None specified                     
  Profile:                 FULL_PROFILE                       
  Float Round Mode:        NEAR                               
  Max Queue Number:        0(0x0)                             
  Queue Min Size:          0(0x0)                             
  Queue Max Size:          0(0x0)                             
  Queue Type:              MULTI                              
  Node:                    2                                  
  Device Type:             CPU                                
  Cache Info:              
    L1:                      32768(0x8000) KB                   
  Chip ID:                 0(0x0)                             
  ASIC Revision:           0(0x0)                             
  Cacheline Size:          64(0x40)                           
  Max Clock Freq. (MHz):   2000                               
  BDFID:                   0                                  
  Internal Node ID:        2                                  
  Compute Unit:            32                                 
  SIMDs per CU:            0                                  
  Shader Engines:          0                                  
  Shader Arrs. per Eng.:   0                                  
  WatchPts on Addr. Ranges:1                                  
  Memory Properties:       
  Features:                None
  Pool Info:               
    Pool 1                   
      Segment:                 GLOBAL; FLAGS: FINE GRAINED        
      Size:                    132110908(0x7dfda3c) KB            
      Allocatable:             TRUE                               
      Alloc Granule:           4KB                                
      Alloc Recommended Granule:4KB                                
      Alloc Alignment:         4KB                                
      Accessible by all:       TRUE                               
    Pool 2                   
      Segment:                 GLOBAL; FLAGS: EXTENDED FINE GRAINED
      Size:                    132110908(0x7dfda3c) KB            
      Allocatable:             TRUE                               
      Alloc Granule:           4KB                                
      Alloc Recommended Granule:4KB                                
      Alloc Alignment:         4KB                                
      Accessible by all:       TRUE                               
    Pool 3                   
      Segment:                 GLOBAL; FLAGS: KERNARG, FINE GRAINED
      Size:                    132110908(0x7dfda3c) KB            
      Allocatable:             TRUE                               
      Alloc Granule:           4KB                                
      Alloc Recommended Granule:4KB                                
      Alloc Alignment:         4KB                                
      Accessible by all:       TRUE                               
    Pool 4                   
      Segment:                 GLOBAL; FLAGS: COARSE GRAINED      
      Size:                    132110908(0x7dfda3c) KB            
      Allocatable:             TRUE                               
      Alloc Granule:           4KB                                
      Alloc Recommended Granule:4KB                                
      Alloc Alignment:         4KB                                
      Accessible by all:       TRUE                               
  ISA Info:                
*******                  
Agent 4                  
*******                  
  Name:                    AMD EPYC 7A53 64-Core Processor    
  Uuid:                    CPU-XX                             
  Marketing Name:          AMD EPYC 7A53 64-Core Processor    
  Vendor Name:             CPU                                
  Feature:                 None specified                     
  Profile:                 FULL_PROFILE                       
  Float Round Mode:        NEAR                               
  Max Queue Number:        0(0x0)                             
  Queue Min Size:          0(0x0)                             
  Queue Max Size:          0(0x0)                             
  Queue Type:              MULTI                              
  Node:                    3                                  
  Device Type:             CPU                                
  Cache Info:              
    L1:                      32768(0x8000) KB                   
  Chip ID:                 0(0x0)                             
  ASIC Revision:           0(0x0)                             
  Cacheline Size:          64(0x40)                           
  Max Clock Freq. (MHz):   2000                               
  BDFID:                   0                                  
  Internal Node ID:        3                                  
  Compute Unit:            32                                 
  SIMDs per CU:            0                                  
  Shader Engines:          0                                  
  Shader Arrs. per Eng.:   0                                  
  WatchPts on Addr. Ranges:1                                  
  Memory Properties:       
  Features:                None
  Pool Info:               
    Pool 1                   
      Segment:                 GLOBAL; FLAGS: FINE GRAINED        
      Size:                    132055844(0x7df0324) KB            
      Allocatable:             TRUE                               
      Alloc Granule:           4KB                                
      Alloc Recommended Granule:4KB                                
      Alloc Alignment:         4KB                                
      Accessible by all:       TRUE                               
    Pool 2                   
      Segment:                 GLOBAL; FLAGS: EXTENDED FINE GRAINED
      Size:                    132055844(0x7df0324) KB            
      Allocatable:             TRUE                               
      Alloc Granule:           4KB                                
      Alloc Recommended Granule:4KB                                
      Alloc Alignment:         4KB                                
      Accessible by all:       TRUE                               
    Pool 3                   
      Segment:                 GLOBAL; FLAGS: KERNARG, FINE GRAINED
      Size:                    132055844(0x7df0324) KB            
      Allocatable:             TRUE                               
      Alloc Granule:           4KB                                
      Alloc Recommended Granule:4KB                                
      Alloc Alignment:         4KB                                
      Accessible by all:       TRUE                               
    Pool 4                   
      Segment:                 GLOBAL; FLAGS: COARSE GRAINED      
      Size:                    132055844(0x7df0324) KB            
      Allocatable:             TRUE                               
      Alloc Granule:           4KB                                
      Alloc Recommended Granule:4KB                                
      Alloc Alignment:         4KB                                
      Accessible by all:       TRUE                               
  ISA Info:                
*******                  
Agent 5                  
*******                  
  Name:                    gfx90a                             
  Uuid:                    GPU-9d4ab50ff31ca655               
  Marketing Name:          AMD Instinct MI250X                
  Vendor Name:             AMD                                
  Feature:                 KERNEL_DISPATCH                    
  Profile:                 BASE_PROFILE                       
  Float Round Mode:        NEAR                               
  Max Queue Number:        128(0x80)                          
  Queue Min Size:          64(0x40)                           
  Queue Max Size:          131072(0x20000)                    
  Queue Type:              MULTI                              
  Node:                    4                                  
  Device Type:             GPU                                
  Cache Info:              
    L1:                      16(0x10) KB                        
    L2:                      8192(0x2000) KB                    
  Chip ID:                 29704(0x7408)                      
  ASIC Revision:           1(0x1)                             
  Cacheline Size:          128(0x80)                          
  Max Clock Freq. (MHz):   1700                               
  BDFID:                   50688                              
  Internal Node ID:        4                                  
  Compute Unit:            110                                
  SIMDs per CU:            4                                  
  Shader Engines:          8                                  
  Shader Arrs. per Eng.:   1                                  
  WatchPts on Addr. Ranges:4                                  
  Coherent Host Access:    TRUE                               
  Memory Properties:       
  Features:                KERNEL_DISPATCH 
  Fast F16 Operation:      TRUE                               
  Wavefront Size:          64(0x40)                           
  Workgroup Max Size:      1024(0x400)                        
  Workgroup Max Size per Dimension:
    x                        1024(0x400)                        
    y                        1024(0x400)                        
    z                        1024(0x400)                        
  Max Waves Per CU:        32(0x20)                           
  Max Work-item Per CU:    2048(0x800)                        
  Grid Max Size:           4294967295(0xffffffff)             
  Grid Max Size per Dimension:
    x                        4294967295(0xffffffff)             
    y                        4294967295(0xffffffff)             
    z                        4294967295(0xffffffff)             
  Max fbarriers/Workgrp:   32                                 
  Packet Processor uCode:: 92                                 
  SDMA engine uCode::      9                                  
  IOMMU Support::          None                               
  Pool Info:               
    Pool 1                   
      Segment:                 GLOBAL; FLAGS: COARSE GRAINED      
      Size:                    67092480(0x3ffc000) KB             
      Allocatable:             TRUE                               
      Alloc Granule:           4KB                                
      Alloc Recommended Granule:2048KB                             
      Alloc Alignment:         4KB                                
      Accessible by all:       FALSE                              
    Pool 2                   
      Segment:                 GLOBAL; FLAGS: EXTENDED FINE GRAINED
      Size:                    67092480(0x3ffc000) KB             
      Allocatable:             TRUE                               
      Alloc Granule:           4KB                                
      Alloc Recommended Granule:2048KB                             
      Alloc Alignment:         4KB                                
      Accessible by all:       FALSE                              
    Pool 3                   
      Segment:                 GLOBAL; FLAGS: FINE GRAINED        
      Size:                    67092480(0x3ffc000) KB             
      Allocatable:             TRUE                               
      Alloc Granule:           4KB                                
      Alloc Recommended Granule:2048KB                             
      Alloc Alignment:         4KB                                
      Accessible by all:       FALSE                              
    Pool 4                   
      Segment:                 GROUP                              
      Size:                    64(0x40) KB                        
      Allocatable:             FALSE                              
      Alloc Granule:           0KB                                
      Alloc Recommended Granule:0KB                                
      Alloc Alignment:         0KB                                
      Accessible by all:       FALSE                              
  ISA Info:                
    ISA 1                    
      Name:                    amdgcn-amd-amdhsa--gfx90a:sramecc+:xnack-
      Machine Models:          HSA_MACHINE_MODEL_LARGE            
      Profiles:                HSA_PROFILE_BASE                   
      Default Rounding Mode:   NEAR                               
      Default Rounding Mode:   NEAR                               
      Fast f16:                TRUE                               
      Workgroup Max Size:      1024(0x400)                        
      Workgroup Max Size per Dimension:
        x                        1024(0x400)                        
        y                        1024(0x400)                        
        z                        1024(0x400)                        
      Grid Max Size:           4294967295(0xffffffff)             
      Grid Max Size per Dimension:
        x                        4294967295(0xffffffff)             
        y                        4294967295(0xffffffff)             
        z                        4294967295(0xffffffff)             
      FBarrier Max Size:       32                                 
*** Done ***             

```

3. Post output of `AMDGPU.versioninfo()` if possible.
```
┌───────────┬──────────────────┬───────────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────
│ Available │ Name             │ Version   │ Path                                                                                                   ⋯
├───────────┼──────────────────┼───────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────
│     +     │ LLD              │ -         │ /opt/rocm-6.3.4/lib/llvm/bin/ld.lld                                                                    ⋯
│     +     │ Device Libraries │ -         │ /projappl/project_462000008/decristoforo/.julia/artifacts/b46ab46ef568406312e5f500efb677511199c2f9/amd ⋯
│     +     │ HIP              │ 6.3.42134 │ /opt/rocm-6.3.4/lib/libamdhip64.so                                                                     ⋯
│     +     │ rocBLAS          │ 4.3.0     │ /opt/rocm-6.3.4/lib/librocblas.so                                                                      ⋯
│     +     │ rocSOLVER        │ 3.27.0    │ /opt/rocm-6.3.4/lib/librocsolver.so                                                                    ⋯
│     +     │ rocSPARSE        │ 3.3.0     │ /opt/rocm-6.3.4/lib/librocsparse.so                                                                    ⋯
│     +     │ rocRAND          │ 2.10.5    │ /opt/rocm-6.3.4/lib/librocrand.so                                                                      ⋯
│     +     │ rocFFT           │ 1.0.31    │ /opt/rocm-6.3.4/lib/librocfft.so                                                                       ⋯
│     +     │ MIOpen           │ 3.3.0     │ /opt/rocm-6.3.4/lib/libMIOpen.so                                                                       ⋯
└───────────┴──────────────────┴───────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────
                                                                                                                                     1 column omitted

[ Info: AMDGPU devices
┌────┬─────────────────────┬────────────────────────┬───────────┬────────────┬───────────────┐
│ Id │                Name │               GCN arch │ Wavefront │     Memory │ Shared Memory │
├────┼─────────────────────┼────────────────────────┼───────────┼────────────┼───────────────┤
│  1 │ AMD Instinct MI250X │ gfx90a:sramecc+:xnack- │        64 │ 63.984 GiB │    64.000 KiB │
└────┴─────────────────────┴────────────────────────┴───────────┴────────────┴───────────────┘


```

## Reproducing the bug

1. Describe what's not working.

`AMDGPU.rocFFT`'s process-global plan cache (`AMDGPU.rocFFT.IDLE_HANDLES`, a `HandleCache` in `src/cache.jl/src/fft/wrappers.jl`) never releases a plan handle unless more than `max_entries` (32) idle handles share the exact same cache key ((`context`, `fft_type`, `dims`, `eltype`, `inplace`, `region`)). A workload that plans many FFTs of distinct shapes (rebuilding FFT plans for a sweep of different problem sizes in one process) accumulates one leaked `rocfft_plan` handle per shape forever, since no shape ever repeats 32+ times. This eventually exhausts GPU memory and a later, unrelated `rocfft_plan_create` call fails with `ROCFFTError(rocfft_status_failure)`.

Note: `cache.jl` has a standing # TODO: https://github.com/JuliaGPU/AMDGPU.jl/blob/16f5974fe0ee0b1b63f249b6592caf5b335b6485/src/cache.jl#L6

2. Provide MWE to reproduce it (if possible).
⚠️ **DISCLAIMER**: AI was involved when writing this scrit.  
```Julia
using AMDGPU

const N_SIZES = length(ARGS) >= 1 ? parse(Int, ARGS[1]) : 6000
const CHECKPOINT_EVERY = 50

# One real->complex forward plan + one complex->real inverse plan for a given length,
# exercised once and then dropped. This is exactly the access pattern
# SpeedyWeather.jl's SpectralTransform uses: one FFT plan per latitude ring (ring lengths
# differ across a Gaussian/octahedral grid), rebuilt from scratch for every distinct
# resolution in a benchmark sweep.
function run_one!(len::Int)
    x = ROCArray(rand(Float32, len))
    p_fwd = AMDGPU.rocFFT.plan_rfft(x, (1,))
    y = p_fwd * x
    p_inv = AMDGPU.rocFFT.plan_brfft(y, len, (1,))
    z = p_inv * y
    AMDGPU.synchronize()
    return nothing
end

function cache_stats()
    idle = AMDGPU.rocFFT.IDLE_HANDLES.idle_handles
    n_keys = length(idle)
    n_idle = isempty(idle) ? 0 : sum(length(v) for v in values(idle))
    return n_keys, n_idle
end

function main(n_sizes::Int)
    println("AMDGPU version: ", AMDGPU.pkgversion(AMDGPU))
    try
        AMDGPU.versioninfo()
    catch err
        println("AMDGPU.versioninfo() failed: ", err)
    end

    free0, total0 = AMDGPU.info()
    println("Initial GPU memory: ", Base.format_bytes(free0), " free / ", Base.format_bytes(total0), " total")
    println("Planning $n_sizes distinct FFT lengths (16, 18, 20, ... in steps of 2)...")

    n_done = 0
    for i in 1:n_sizes
        len = 16 + 2 * (i - 1)
        try
            run_one!(len)
        catch err
            bt = catch_backtrace()
            println()
            println("=== REPRODUCED after $i distinct plan sizes (last size = $len) ===")
            n_keys, n_idle = cache_stats()
            free, total = AMDGPU.info()
            println("cached_plan_shapes=$n_keys idle_handles=$n_idle free=$(Base.format_bytes(free))")
            showerror(stdout, err, bt)
            println()
            println("RESULT: FAIL (reproduced) after $i distinct FFT plan sizes")
            exit(1)
        end
        n_done = i

        if i % CHECKPOINT_EVERY == 0
            GC.gc()               # force finalizers (release_plan!) to run
            AMDGPU.synchronize()
            free, total = AMDGPU.info()
            n_keys, n_idle = cache_stats()
            println(
                "[$i/$n_sizes] free=$(Base.format_bytes(free)) " *
                    "cached_plan_shapes=$n_keys idle_handles=$n_idle"
            )
        end
    end

    free_end, _ = AMDGPU.info()
    println()
    println(
        "RESULT: did not reproduce with n_sizes=$n_sizes (free memory dropped from ",
        Base.format_bytes(free0), " to ", Base.format_bytes(free_end),
        "). Try a larger n_sizes argument, or run on a GPU with less free memory."
    )
    return
end

main(N_SIZES)

```

On LUMI this results in the following error:
```Julia
julia> main(N_SIZES)
AMDGPU version: 2.8.0
AMDGPU versioninfo
AMDGPU.versioninfo() failed: ErrorException("could not load symbol \"hiptensorGetVersion\":\n/opt/rocm-6.3.4/lib/libhiptensor.so: undefined symbol: hiptensorGetVersion")
Initial GPU memory: 63.896 GiB free / 63.984 GiB total
Planning 6000 distinct FFT lengths (16, 18, 20, ... in steps of 2)...
[50/6000] free=63.168 GiB cached_plan_shapes=100 idle_handles=100
[100/6000] free=62.385 GiB cached_plan_shapes=200 idle_handles=200
[150/6000] free=61.555 GiB cached_plan_shapes=300 idle_handles=300
[200/6000] free=60.664 GiB cached_plan_shapes=400 idle_handles=400
[250/6000] free=59.768 GiB cached_plan_shapes=500 idle_handles=500
[300/6000] free=58.758 GiB cached_plan_shapes=600 idle_handles=600
[350/6000] free=57.760 GiB cached_plan_shapes=700 idle_handles=700
[400/6000] free=56.738 GiB cached_plan_shapes=800 idle_handles=800
[450/6000] free=55.705 GiB cached_plan_shapes=900 idle_handles=900
[500/6000] free=54.672 GiB cached_plan_shapes=1000 idle_handles=1000
[550/6000] free=53.580 GiB cached_plan_shapes=1100 idle_handles=1100
[600/6000] free=52.500 GiB cached_plan_shapes=1200 idle_handles=1200
[650/6000] free=51.396 GiB cached_plan_shapes=1300 idle_handles=1300
[700/6000] free=50.309 GiB cached_plan_shapes=1400 idle_handles=1400
[750/6000] free=49.213 GiB cached_plan_shapes=1500 idle_handles=1500
[800/6000] free=48.094 GiB cached_plan_shapes=1600 idle_handles=1600
[850/6000] free=47.000 GiB cached_plan_shapes=1700 idle_handles=1700
[900/6000] free=45.904 GiB cached_plan_shapes=1800 idle_handles=1800
[950/6000] free=44.785 GiB cached_plan_shapes=1900 idle_handles=1900
[1000/6000] free=43.674 GiB cached_plan_shapes=2000 idle_handles=2000
[1050/6000] free=42.561 GiB cached_plan_shapes=2100 idle_handles=2100
[1100/6000] free=41.449 GiB cached_plan_shapes=2200 idle_handles=2200
[1150/6000] free=40.338 GiB cached_plan_shapes=2300 idle_handles=2300
[1200/6000] free=39.211 GiB cached_plan_shapes=2400 idle_handles=2400
[1250/6000] free=38.092 GiB cached_plan_shapes=2500 idle_handles=2500
[1300/6000] free=36.973 GiB cached_plan_shapes=2600 idle_handles=2600
[1350/6000] free=35.861 GiB cached_plan_shapes=2700 idle_handles=2700
[1400/6000] free=34.750 GiB cached_plan_shapes=2800 idle_handles=2800
[1450/6000] free=33.631 GiB cached_plan_shapes=2900 idle_handles=2900
[1500/6000] free=32.496 GiB cached_plan_shapes=3000 idle_handles=3000
[1550/6000] free=31.367 GiB cached_plan_shapes=3100 idle_handles=3100
[1600/6000] free=30.240 GiB cached_plan_shapes=3200 idle_handles=3200
[1650/6000] free=29.113 GiB cached_plan_shapes=3300 idle_handles=3300
[1700/6000] free=28.002 GiB cached_plan_shapes=3400 idle_handles=3400
[1750/6000] free=26.875 GiB cached_plan_shapes=3500 idle_handles=3500
[1800/6000] free=25.756 GiB cached_plan_shapes=3600 idle_handles=3600
[1850/6000] free=24.613 GiB cached_plan_shapes=3700 idle_handles=3700
[1900/6000] free=23.486 GiB cached_plan_shapes=3800 idle_handles=3800
[1950/6000] free=22.359 GiB cached_plan_shapes=3900 idle_handles=3900
[2000/6000] free=21.225 GiB cached_plan_shapes=4000 idle_handles=4000
[2050/6000] free=19.939 GiB cached_plan_shapes=4100 idle_handles=4100
[2100/6000] free=17.883 GiB cached_plan_shapes=4200 idle_handles=4200
[2150/6000] free=15.824 GiB cached_plan_shapes=4300 idle_handles=4300
[2200/6000] free=13.848 GiB cached_plan_shapes=4400 idle_handles=4400
[2250/6000] free=11.781 GiB cached_plan_shapes=4500 idle_handles=4500
[2300/6000] free=9.734 GiB cached_plan_shapes=4600 idle_handles=4600
[2350/6000] free=7.656 GiB cached_plan_shapes=4700 idle_handles=4700
[2400/6000] free=5.615 GiB cached_plan_shapes=4800 idle_handles=4800
[2450/6000] free=3.584 GiB cached_plan_shapes=4900 idle_handles=4900
[2500/6000] free=1.533 GiB cached_plan_shapes=5000 idle_handles=5000

=== REPRODUCED after 2524 distinct plan sizes (last size = 5062) ===
cached_plan_shapes=5033 idle_handles=5033 free=576.000 MiB
ROCFFTError(code rocfft_status_failure, an operation failed)
Stacktrace:
  [1] check
    @ /projappl/project_462000008/decristoforo/.julia/packages/AMDGPU/HUKS4/src/fft/error.jl:38 [inlined]
  [2] macro expansion
    @ /projappl/project_462000008/decristoforo/.julia/packages/AMDGPU/HUKS4/src/utils.jl:239 [inlined]
  [3] rocfft_plan_create(plan::Base.RefValue{Ptr{AMDGPU.rocFFT.rocfft_plan_t}}, placement::AMDGPU.rocFFT.rocfft_result_placement_e, transform_type::AMDGPU.rocFFT.rocfft_transform_type_e, precision::AMDGPU.rocFFT.rocfft_precision_e, dimensions::Int64, lengths::Vector{Int64}, number_of_transforms::Int64, description::Ptr{Nothing})
    @ AMDGPU.rocFFT /projappl/project_462000008/decristoforo/.julia/packages/AMDGPU/HUKS4/src/fft/librocfft.jl:92
  [4] create_plan(xtype::AMDGPU.rocFFT.rocfft_transform_type_e, xdims::Tuple{Int64}, T::Type, inplace::Bool, region::Tuple{Int64})
    @ AMDGPU.rocFFT /projappl/project_462000008/decristoforo/.julia/packages/AMDGPU/HUKS4/src/fft/wrappers.jl:40
  [5] #get_plan##0
    @ /projappl/project_462000008/decristoforo/.julia/packages/AMDGPU/HUKS4/src/fft/wrappers.jl:11 [inlined]
  [6] pop!(f::AMDGPU.rocFFT.var"#get_plan##0#get_plan##1"{AMDGPU.rocFFT.rocfft_transform_type_e, Tuple{Int64}, Type{Float32}, Bool, Tuple{Int64}}, cache::HandleCache{Tuple{HIPContext, AMDGPU.rocFFT.rocfft_transform_type_e, NTuple{N, Int64} where N, Type, Bool, Any}, Tuple{Ptr{AMDGPU.rocFFT.rocfft_plan_t}, Int64}}, key::Tuple{HIPContext, AMDGPU.rocFFT.rocfft_transform_type_e, Tuple{Int64}, DataType, Bool, Tuple{Int64}})
    @ AMDGPU /projappl/project_462000008/decristoforo/.julia/packages/AMDGPU/HUKS4/src/cache.jl:49
  [7] get_plan
    @ /projappl/project_462000008/decristoforo/.julia/packages/AMDGPU/HUKS4/src/fft/wrappers.jl:11 [inlined]
  [8] plan_rfft(X::ROCArray{Float32, 1, AMDGPU.Runtime.Mem.HIPBuffer}, region::Tuple{Int64})
    @ AMDGPU.rocFFT /projappl/project_462000008/decristoforo/.julia/packages/AMDGPU/HUKS4/src/fft/fft.jl:149
  [9] run_one!(len::Int64)
    @ Main ./REPL[4]:8
 [10] main(n_sizes::Int64)
    @ Main ./REPL[6]:17
 [11] top-level scope
    @ REPL[7]:1
 [12] __repl_entry_eval_expanded_with_loc(mod::Module, ast::Any, toplevel_file::Ref{Ptr{UInt8}}, toplevel_line::Ref{Int32})
    @ REPL /pfs/lustrep4/users/decristoforo/.julia/juliaup/julia-1.12.7+0.x64.linux.gnu/share/julia/stdlib/v1.12/REPL/src/REPL.jl:301
 [13] toplevel_eval_with_hooks(mod::Module, ast::Any, toplevel_file::Any, toplevel_line::Any)
    @ REPL /pfs/lustrep4/users/decristoforo/.julia/juliaup/julia-1.12.7+0.x64.linux.gnu/share/julia/stdlib/v1.12/REPL/src/REPL.jl:308
 [14] toplevel_eval_with_hooks(mod::Module, ast::Any, toplevel_file::Any, toplevel_line::Any)
    @ REPL /pfs/lustrep4/users/decristoforo/.julia/juliaup/julia-1.12.7+0.x64.linux.gnu/share/julia/stdlib/v1.12/REPL/src/REPL.jl:312
 [15] toplevel_eval_with_hooks
    @ /pfs/lustrep4/users/decristoforo/.julia/juliaup/julia-1.12.7+0.x64.linux.gnu/share/julia/stdlib/v1.12/REPL/src/REPL.jl:305 [inlined]
 [16] eval_user_input(ast::Any, backend::REPL.REPLBackend, mod::Module)
    @ REPL /pfs/lustrep4/users/decristoforo/.julia/juliaup/julia-1.12.7+0.x64.linux.gnu/share/julia/stdlib/v1.12/REPL/src/REPL.jl:330
 [17] repl_backend_loop(backend::REPL.REPLBackend, get_module::Function)
    @ REPL /pfs/lustrep4/users/decristoforo/.julia/juliaup/julia-1.12.7+0.x64.linux.gnu/share/julia/stdlib/v1.12/REPL/src/REPL.jl:452
 [18] start_repl_backend(backend::REPL.REPLBackend, consumer::Any; get_module::Function)
    @ REPL /pfs/lustrep4/users/decristoforo/.julia/juliaup/julia-1.12.7+0.x64.linux.gnu/share/julia/stdlib/v1.12/REPL/src/REPL.jl:427
 [19] start_repl_backend
    @ /pfs/lustrep4/users/decristoforo/.julia/juliaup/julia-1.12.7+0.x64.linux.gnu/share/julia/stdlib/v1.12/REPL/src/REPL.jl:424 [inlined]
 [20] run_repl(repl::REPL.AbstractREPL, consumer::Any; backend_on_current_task::Bool, backend::Any)
    @ REPL /pfs/lustrep4/users/decristoforo/.julia/juliaup/julia-1.12.7+0.x64.linux.gnu/share/julia/stdlib/v1.12/REPL/src/REPL.jl:653
 [21] run_repl(repl::REPL.AbstractREPL, consumer::Any)
    @ REPL /pfs/lustrep4/users/decristoforo/.julia/juliaup/julia-1.12.7+0.x64.linux.gnu/share/julia/stdlib/v1.12/REPL/src/REPL.jl:639
 [22] run_std_repl(REPL::Module, quiet::Bool, banner::Symbol, history_file::Bool)
    @ Base ./client.jl:478
 [23] run_main_repl(interactive::Bool, quiet::Bool, banner::Symbol, history_file::Bool)
    @ Base ./client.jl:499
 [24] repl_main
    @ ./client.jl:586 [inlined]
 [25] _start()
    @ Base ./client.jl:561
RESULT: FAIL (reproduced) after 2524 distinct FFT plan sizes
```

gregordecristoforo:
Proposed fix #1053 

`HandleCache` got a `max_idle` budget that bounds the total number of idle handles across all keys (not just per-key via `max_entries`), with a new `idle_dtors` dict holding each idle handle's destructor so that `_evict_idle!` can reclaim handles for any key once the budget is exceeded, running those destructors outside the lock. Also, `release_plan!` was changed to capture the plan handle in a local variable and close over that (`() -> rocfft_plan_destroy(handle)`) instead of closing over the plan and reading `plan.handle` lazily.

With this implementation the MWE in #1053 gives the following output on LUMI:

```shell
➜  AMDGPU.jl git:(gd/plan-handle-cache-fix) ✗ julia --project=. mwe.jl
Initial GPU memory: 63.896 GiB free / 63.984 GiB total
Planning 6000 distinct FFT lengths (16, 18, 20, ... in steps of 2)...
[50/6000] free=63.371 GiB cached_plan_shapes=64 idle_handles=64
used=0 bytes cached(pool reserved)=464 bytes
[100/6000] free=63.293 GiB cached_plan_shapes=64 idle_handles=64
used=0 bytes cached(pool reserved)=864 bytes
[150/6000] free=63.293 GiB cached_plan_shapes=64 idle_handles=64
used=0 bytes cached(pool reserved)=1.234 KiB
[200/6000] free=63.262 GiB cached_plan_shapes=64 idle_handles=64
used=0 bytes cached(pool reserved)=1.625 KiB
...
```

Is this PR going into the right direction?


gregordecristoforo:
@luraess Would you be able to have a look?

luraess:
Thanks. I checked the proposed changes and iterated on it with the robot. The direction looks right to me (diagnosis matches what I tested, the fix works).

I ran it on MI300 (gfx942, ROCm 7.2.3). The #1053 MWE now holds at 64 idle handles. Free memory goes flat from iteration 100 through 1200, with 195 MiB of drift that plateaus. Your tests pass as well: core 104/104, fft 7/7, `hip_rocarray/fft` at 268/268. The leak in your test measures 1.02 MiB against the 400 MiB budget.

A few things I noticed.

**On framing.** Capturing the handle by value in `release_plan!` looks like a prerequisite of this PR rather than a bug fix. Before it, the destructor always ran synchronously inside `push!`. That happened ahead of `unsafe_free!` setting `plan.handle = C_NULL`, so nothing shipped broken. Might be worth rewording the comment and the test's "used to queue…". There is a second argument for the change though: closing over `plan` and parking it in a process-global dict would resurrect a finalized object.

**1. Eviction order in `_evict_idle!`.** This is the one I'd look at before merging.

Your docstring already notes it is "a count budget, not an LRU". It may be a little sharper than arbitrary, though. Eviction restarts from the front each time and drains until the budget is met, so it tends to hit the same keys repeatedly. I measured this against the real `IDLE_HANDLES`, over 200 iterations of "N shapes reused every iteration, plus M shapes used once":

```
max_idle = 64, max_entries = 32
hot=1   cold/iter=4  ->   77 rebuilds of a hot plan (ideal 1)
hot=4   cold/iter=4  ->  800 rebuilds (ideal 4)    <- 100% miss
hot=8   cold/iter=8  -> 1547 rebuilds (ideal 8)
hot=16  cold/iter=4  -> 3200 rebuilds (ideal 16)   <- 100% miss
```

On `main` all four rows sit at `ideal`, since nothing is evicted at all. A rebuild costs about 4.8 ms when rocFFT's RTC cache is warm for that shape, and 14.4 ms when it is cold. At four reused shapes, every lookup misses.

This might be one change together with `idle_dtors` rather than two. A single `Dict{K, Vector{IdleHandle{V}}}`, with entries carrying `(handle, dtor, seq)`, would collapse the two dicts into one and make the parallel invariant unrepresentable. It also gives eviction a global order to work from. Evicting the lowest `seq` brings all four rows back to `ideal`. Each key's vector is already in ascending `seq` order, so only its front is a candidate. That keeps the scan at O(number of keys), and the keys are bounded by `max_idle + 1`.

<details>
<summary>Diff, if useful (<code>src/cache.jl</code> only, and it takes <code>max_idle = 64</code> as given since that is the separate question in 2)</summary>

```diff
diff --git a/src/cache.jl b/src/cache.jl
index bfc41a0d..4884ba02 100644
--- a/src/cache.jl
+++ b/src/cache.jl
@@ -7,13 +7,17 @@
 
 export HandleCache
 
+# One idle handle, together with the destructor that reclaims it and the global
+# insertion order used to pick an eviction victim.
+struct IdleHandle{V}
+    handle::V
+    dtor::Any
+    seq::Int
+end
+
 struct HandleCache{K, V}
     active_handles::Set{Pair{K, V}}
-    idle_handles::Dict{K, Vector{V}}
-    # Destructor closure for every idle handle, kept parallel to `idle_handles`
-    # so that idle handles for *any* key can be reclaimed (not just the one
-    # currently being pushed).
-    idle_dtors::Dict{K, Vector{Any}}
+    idle_handles::Dict{K, Vector{IdleHandle{V}}}
     lock::Base.ThreadSynchronizer
     # TODO when finalizers are run on their own tasks use reentrant lock
 
@@ -23,56 +27,59 @@ struct HandleCache{K, V}
     # cache for workloads that use a large number of distinct keys (e.g. rocFFT
     # plans for many different shapes), which would otherwise leak one handle
     # per key forever since no single key ever reaches `max_entries`. See #1053.
-    # Inert for caches keyed on something with few values (e.g. `HIPContext`),
-    # where the per-key `max_entries` budget is the one that matters.
     max_idle::Int
+    # Stamps each idle handle so `_evict_idle!` can evict in least-recently-cached
+    # order across all keys, instead of whichever key `Dict` iteration visits first.
+    seq::Base.RefValue{Int}
 
     function HandleCache{K, V}(max_entries::Int = 32, max_idle::Int = 64) where {K, V}
         new{K,V}(
             Set{Pair{K, V}}(),
-            Dict{K, Vector{V}}(),
-            Dict{K, Vector{Any}}(),
+            Dict{K, Vector{IdleHandle{V}}}(),
             Base.ThreadSynchronizer(),
-            max_entries, max_idle)
+            max_entries, max_idle, Ref(0))
     end
 end
 
+# Total number of idle handles across all keys.
+# Must be called while holding `cache.lock`.
+total_idle(cache::HandleCache) = sum(length, values(cache.idle_handles); init = 0)
+
 # Take an idle handle for `key` out of the cache, or `nothing` if there is none.
+# Its destructor is dropped: the caller owns the handle again, and hands us a
+# fresh destructor when it comes back through `push!`.
 # Must be called while holding `cache.lock`.
 function _take_idle!(cache::HandleCache{K, V}, key) where {K, V}
-    handles = get(cache.idle_handles, key, nothing)
-    (handles ≡ nothing || isempty(handles)) && return nothing
-    handle = pop!(handles)
-    pop!(cache.idle_dtors[key])
-    if isempty(handles)
-        delete!(cache.idle_handles, key)
-        delete!(cache.idle_dtors, key)
-    end
-    return handle
+    entries = get(cache.idle_handles, key, nothing)
+    entries ≡ nothing && return nothing
+    entry = pop!(entries)
+    isempty(entries) && delete!(cache.idle_handles, key)
+    return entry.handle
 end
 
 # Evict idle handles until at most `max_idle` remain across all keys, returning
-# the destructor closures of the evicted handles to be run by the caller outside
-# the lock. This is a count budget, not an LRU: keys are visited in `Dict`
-# iteration order (arbitrary after deletions) and each key's oldest handles go
-# first. Must be called while holding `cache.lock`.
+# the destructors of the evicted handles to be run by the caller outside the
+# lock. Victims are chosen in least-recently-cached order across all keys, so a
+# key that keeps being reused survives churn from keys that are used once.
+# Each key's vector is in ascending `seq` order, so only its front is a
+# candidate, making the scan O(number of keys) per eviction.
+# Must be called while holding `cache.lock`.
 function _evict_idle!(cache::HandleCache{K, V}) where {K, V}
     evicted = Any[]
-    total = sum(length, values(cache.idle_handles); init = 0)
-    total ≤ cache.max_idle && return evicted
-    for key in collect(keys(cache.idle_handles))
-        handles = cache.idle_handles[key]
-        dtors = cache.idle_dtors[key]
-        while !isempty(handles) && total > cache.max_idle
-            popfirst!(handles)
-            push!(evicted, popfirst!(dtors))
-            total -= 1
-        end
-        if isempty(handles)
-            delete!(cache.idle_handles, key)
-            delete!(cache.idle_dtors, key)
+    total = total_idle(cache)
+    while total > cache.max_idle
+        victim, oldest = nothing, typemax(Int)
+        for (key, entries) in cache.idle_handles
+            if entries[1].seq < oldest
+                victim, oldest = key, entries[1].seq
+            end
         end
-        total ≤ cache.max_idle && break
+        victim ≡ nothing && break
+
+        entries = cache.idle_handles[victim]
+        push!(evicted, popfirst!(entries).dtor)
+        isempty(entries) && delete!(cache.idle_handles, victim)
+        total -= 1
     end
     return evicted
 end
@@ -109,13 +116,16 @@ function Base.push!(f::Function, cache::HandleCache{K, V}, key::K, handle::V) wh
             """)
         delete!(cache.active_handles, key => handle)
 
-        handles = get!(() -> V[], cache.idle_handles, key)
+        entries = get(cache.idle_handles, key, nothing)
         # Keep the original semantics: save while the key holds ≤ `max_entries`,
         # so a key can grow to `max_entries + 1`.
-        saved = length(handles) ≤ cache.max_entries
+        saved = (entries ≡ nothing ? 0 : length(entries)) ≤ cache.max_entries
         if saved
-            push!(handles, handle)
-            push!(get!(() -> Any[], cache.idle_dtors, key), f)
+            # Only materialise the vector once we know we are filling it, so a
+            # key is present in `idle_handles` iff it holds at least one handle.
+            entries ≡ nothing &&
+                (entries = cache.idle_handles[key] = IdleHandle{V}[])
+            push!(entries, IdleHandle{V}(handle, f, (cache.seq[] += 1)))
         end
 
         # Enforce the global idle-handle budget.
```

Checked on MI300: invariants hold over 200 randomised trials x 400 ops (`max_entries` 0-5, `max_idle` 0-12, with no lost or double-freed handles, the budget never exceeded, and no empty vectors left behind). The table above becomes 1 / 3 / 4 / 8 rebuilds against the real `IDLE_HANDLES`, and `hip_rocarray/fft` stays green.

</details>

The destructor's context could then be a two-liner in `release_plan!`. An evicted plan may now be destroyed under whichever context happens to be current, where before it ran inline under the one that built the key. `library_state` wraps its destructor in `context!(tls.context)`, presumably for this reason:

```julia
ctx, handle = AMDGPU.context(), plan.handle
push!(() -> AMDGPU.context!(() -> rocfft_plan_destroy(handle), ctx), IDLE_HANDLES, key, value)
```

There is a fuller version: one `dtor(key, handle)` on the cache, with no per-handle closures, which is roughly the TODO you removed. It touches all six construction sites though, so probably not this PR.

**2. `max_idle` and the `HIPContext` caches.** The comment says it is inert for those, but I don't think it is. `max_entries` allows 33 idle handles per key, so two contexts already exceed 64. I hit that on a 2-GCD allocation. rocBLAS, rocSPARSE, MIOpen, hipTENSOR and rocRAND all share this default, and a rocBLAS handle is more expensive to rebuild than an FFT plan. One option would be `max_idle = typemax(Int)` as the default, with an explicit budget only on rocFFT's `IDLE_HANDLES`.

Related: a count may be a weak proxy for what is actually being bounded. #1053's numbers work out to roughly 10 MB per plan. The small 1-D plans here are nearer 3 MB, and 64 large 3-D plans would be different again. Making the number settable could help until the memory-pressure TODO lands.

**3. Tests.**

- The closure introspection asserts a local variable name. Renaming `handle` to `h` in `release_plan!` makes `fieldnames` return `(:h,)`, so the test fails with nothing else changed. The leak subtest below it seems to cover the same ground, so I'd probably drop it.
- `churn(200:2:598)` costs about 66 s. Given the margin is 1.02 MiB against 400 MiB, `churn(200:2:400)` should give the same signal in half the time.
- Once eviction has a defined order, a test pinning it would be useful: a hot key surviving cold churn. That is the part that can regress quietly.
- `sum(length, values(...); init = 0)` appears three times. A `total_idle(cache)` helper would DRY it, and give the tests an accessor instead of reaching into fields.

**Small one:** the `try`/`catch` now swallows a failing `rocfft_plan_destroy` that used to throw. That seems reasonable for the deferred case. Worth noting that `@error` allocates and takes the logging lock, and `push!` is reached from a finalizer, so `@debug` or a rate limit might be safer.

Separately, and pre-existing: `release_plan!` builds its key from `AMDGPU.context()` at finalization time, not the context the plan was created in. On multi-GPU that could already file a plan under the wrong device. Probably its own issue rather than something for this PR.

gregordecristoforo:
Thanks @luraess for the detailed feedback! I'll probably have all day on Tuesday next week to implement it. 

> Separately, and pre-existing: release_plan! builds its key from AMDGPU.context() at finalization time, not the context the plan was created in. On multi-GPU that could already file a plan under the wrong device. Probably its own issue rather than something for this PR.

I am happy to look at this in a separate PR. I can use LUMI for multi-GPU testing.

gregordecristoforo:
I had a look at your feedback @luraess. Here is an overview of the changes I made:

1. **Reworded the byval-capture comments**: the `release_plan!` comment and its regression test now describe the current requirement concisely instead of narrating the old buggy behavior.
2. **Fixed eviction order**: collapsed `idle_handles/idle_dtors` into a single `Dict{K, Vector{IdleHandle{V}}}` keyed with a global `seq` counter, so `_evict_idle!` now evicts the globally-oldest idle handle instead of whichever key Dict iteration visits first.
3. **Added a randomized invariant test**: 200 trials of 400 random `pop!/push!` operations against `HandleCache{Int,Int}`, checking after every operation that no handle is lost or destroyed twice, budgets are respected, and no empty vectors are left behind.
4. **Fixed the max_idle default**: `HandleCache` now defaults to an effectively unbounded global idle budget, with a real one (32, 64) set explicitly only on` rocFFT's IDLE_HANDLES`, so the context-keyed caches (`rocBLAS, rocSPARSE, MIOpen, hipTENSOR, rocRAND`) no longer share an inappropriately small global cap.
5. **Fixed the eviction context**: `release_plan!` now wraps the deferred plan destructor in `AMDGPU.context!(..., ctx)` so an evicted plan is destroyed under the context it was created in, not whatever context happens to be current at eviction time.
6. **Cleaned up and extended the FFT cache tests**: dropped a fragile closure-introspection test, shrank an oversized churn range, deduplicated three inline idle-count calculations into a `total_idle` helper, and added a test pinning that a frequently-reused shape survives churn from many one-off shapes.
7. **Softened destructor-failure logging**: swapped `@error` for `@debug` when a cached handle's destructor throws, since that code can run from a finalizer where taking the logging lock and allocating unconditionally is best avoided.

Is there anything else left to do?

luraess:
Looks like all test fail on `bookkeeping stays bounded and device memory does not leak` one.

gregordecristoforo:
I loosened the leak-size threshold in the failing test I added. The exact size seems to vary by GPU/ROCm version.

luraess:
Could this size be queried for the target GPU arch?

gregordecristoforo:
I am not aware of some kind of rocFFT API to query "how much device memory will this kernel's compiled code occupy for arch X" ahead of time (please correct me if I am missing something).

Should I remove this test as it has this arbitrary threshold? An alternative test idea could be to run the same FFT shapes through the cache twice. The first pass pays rocFFT's unpredictable, GPU-arch-dependent one-time kernel-compile cost, and the second pass reuses those already-compiled kernels, so any device memory not recovered there is unambiguously a `HandleCache` leak rather than rocFFT's own cache growth. That would let us assert a small, arch-independent bound (8 MiB or so) on the second pass instead of guessing a "per-plan MiB" constant that varies by GPU/ROCm version.

luraess:
Thanks, this addresses everything from my previous review.

On the leak test: I don't think the variance is mainly the GPU arch. A second pass may read the same noisy counters (tested again on the CI GPU), so it wouldn't help. I'd drop the memory assertion (or just `@info` the value); the bookkeeping and hot-key tests cover the eviction logic. To still guard the by-value handle capture, a deterministic option is to count plans created and plans destroyed with a non-null handle, and check `Δcreated == Δdestroyed + Δidle` across the churn.

Minor:
- The hot-key test takes up to 4.5 min on CI. The same check on a `HandleCache{Int,Int}` in the core tests is instant, e.g.:
```
@testset "HandleCache: hot key survives cold churn" begin
    cache = HandleCache{Int, Int}(32, 8)
    created = Ref(0)
    get_put(key) = (h = pop!(() -> (created[] += 1), cache, key);
                    push!(() -> nothing, cache, key, h); h)
    get_put(0)                    # hot key
    before = created[]
    cold = 1
    for _ in 1:50
        get_put(0)
        for _ in 1:4
            get_put(cold); cold += 1
        end
    end
    # each cold key creates one handle; anything more means the hot key was rebuilt
    @test created[] - before == cold - 1
end
```

- The randomized test adds ~640k `@test`s and has no fixed seed. One `@test` per trial and a seeded RNG would keep failures reproducible.
- The `release_plan!` comment says plans are destroyed under the context they were built under, but `ctx` is the one current at release.

gregordecristoforo:
Thanks for the feedback. I addressed your points in [c34d78e](https://github.com/JuliaGPU/AMDGPU.jl/pull/1070/commits/c34d78e7d74bd39fc6845e44e92c4dd71291a21e).

luraess:
LGTM - thanks! Maybe if you want to give it a final pass and trim some of the inline comments in order to keep the code concise and avoid too much comment lines? I will run a final check on MI300A now that the cluster is back and report.

gregordecristoforo:
I trimmed the comments, please let me know if there is anything else.

luraess:
Thanks!

PR #1071, opened 2026-09-09
Adapt to the library-based AMDGPU_LLVM_Backend_jll 23 and LLVMDowngrader_jll 0.10

maleadt:
JuliaPackaging/Yggdrasil#14727 replaced the `llc`, `lld` and `llvm-downgrade` executables in the GPU LLVM JLLs with shared libraries exposing a small C API, moving the back-ends to LLVM 23 at the same time. JuliaGPU/GPUCompiler.jl#930 makes GPUCompiler call those libraries in-process, and bumps its compat to the new JLL majors. Since this package pins the same JLL, it needs its compat bumped in lockstep, which is what this PR does.

AMDGPU.jl used two of the removed tools itself, so besides the compat bumps (`AMDGPU_LLVM_Backend_jll` → 23, `LLVMDowngrader_jll` → 0.10, `GPUCompiler` → 2.7):

- Linking the kernel object into an HSA code object goes through `AMDGPULink` from `libamdgpu` (`ld.lld -flavor gnu -shared` in-process, no temporary files) whenever the artifact is used. A system `ld.lld` from ROCm is still preferred when found, and still spawned as before. `lld_artifact` now means "link through the library" and no longer comes with a tool path; `versioninfo` shows the library instead.
- Downgrading the device libraries for the in-process LLVM goes through `LLVMDGDowngrade`, one buffer at a time, with the downgrader's error message in the warning when a library is skipped.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -49,9 +49,6 @@ AMDGPUEnzymeCoreExt = "EnzymeCore"
 AMDGPUSparseMatricesCSRExt = "SparseMatricesCSR"
 AMDGPUSpecialFunctionsExt = "SpecialFunctions"
 
-[sources]
-GPUCompiler = {url = "https://github.com/JuliaGPU/GPUCompiler.jl", rev = "tb/jll-libraries"}
-
 [compat]
 AMDGPU_LLVM_Backend_jll = "23"
 AbstractFFTs = "1.0"
--- a/Project.toml
+++ b/Project.toml
@@ -49,8 +49,11 @@ AMDGPUEnzymeCoreExt = "EnzymeCore"
 AMDGPUSparseMatricesCSRExt = "SparseMatricesCSR"
 AMDGPUSpecialFunctionsExt = "SpecialFunctions"
 
+[sources]
+GPUCompiler = {url = "https://github.com/JuliaGPU/GPUCompiler.jl", rev = "tb/jll-libraries"}
+
 [compat]
-AMDGPU_LLVM_Backend_jll = "22.1.8"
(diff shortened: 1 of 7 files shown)


simeonschaub:
What's the reason we need to use `dlopen` and `dlsym` here for calling these libraries? Since they're just regular jlls, can't we use e.g. `libamdgpu` in ccall directly?

maleadt:
> What's the reason we need to use `dlopen` and `dlsym` here for calling these libraries? Since they're just regular jlls, can't we use e.g. `libamdgpu` in ccall directly?

Ah, yes. This was modeled after how we did things in https://github.com/JuliaGPU/GPUCompiler.jl/pull/930, where the names are constructed more dynamically. I'll make it a plain `ccall` here.

maleadt:
Previous CI run was all green, an latest change was NFC, so merging this a little more quickly such that we get GPUCompiler reverse CI back.

luraess:
After merging this into main, we're getting an error on `hip_rocarray/multi_gpu` -> `device switching` https://github.com/JuliaGPU/AMDGPU.jl/blob/63c6c40413001d97eeebb7291911f577e2e04ab3/test/hip_rocarray/multi_gpu.jl#L10 test.

CI log: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/3266148763123078/6677263153381401/-/jobs/16452799405

maleadt:
I don't have the hardware to debug this on, but it's probably a LLVM 23 regression? Claude is pretty good at diagnosing these, and adding backports to our LLVM build.

luraess:
Will look into it.

PR #1072, opened 2026-09-11
Adapt to LLVMDowngrader_jll 0.11 and GPUCompiler 2.8

maleadt:
Bumps the compat bounds to LLVMDowngrader_jll 0.11 (JuliaPackaging/Yggdrasil#14758, built against LLVM 23) and GPUCompiler 2.8 (JuliaGPU/GPUCompiler.jl#931). The downgrader C API is unchanged, so this is compat-only.


christiangnrd:
Close #1080

luraess:
Can you rebase on main to see that CI passes and we could merge

PR #1073, opened 2026-09-14
Avoid boxing compiler jobs during kernel resolution

maleadt:
Kernel resolution constructs a `CompilerJob` immediately before passing it through an unspecialized `compile_or_lookup` boundary, which can force the job onto the heap. Allow specialization on the job's target and compiler-parameter types, and keep the method `@noinline` so its body is not duplicated into callers specialized per kernel. Kernel identity remains a field of the job, so this does not introduce per-kernel specialization or another cache.

Applies the approach from [KernelAbstractions.jl#769](https://github.com/JuliaGPU/KernelAbstractions.jl/pull/769).

Syntax and whitespace checks passed, and the package loaded on Julia 1.13. Device execution and allocation measurements were unavailable because the local HIP runtime was not functional.


luraess:
Again Int128 failing test, on 1.12 only https://github.com/JuliaGPU/AMDGPU.jl/blob/63c6c40413001d97eeebb7291911f577e2e04ab3/test/core/rocarray_broadcast.jl#L59

luraess:
Can you rebase on main to see CI passes then we could merge

PR #1075, opened 2026-09-16
Define KernelState before the device code that references it

gbaraldi:
Root fix for JuliaGPU/AcceleratedKernels.jl#123: on Julia 1.13, `sort!` (and any bounds-checked kernel) fails with `InvalidIRError: unsupported dynamic function invocation (call to kernel_state())` — but only on RDNA2-class GPUs.

**Cause.** Generators run in their defining world (Julia 1.12+). `@generated kernel_state()` (`device/runtime.jl`) references `AMDGPU.KernelState`, which is defined *after* `include("device/Device.jl")`. During AMDGPU's **own precompilation** the intra-package world ordering is live, so the generator throws `UndefVarError: KernelState not defined in AMDGPU` — which `Compiler.get_staged` silently swallows ("user code might throw errors – ignore them"), making inference type `kernel_state()::Any`. The runtime library is then compiled with boxed dynamic dispatch (`jl_apply_generic` on `getproperty`, a `jl_f_throw_methoderror` guard for `signal_exception!(::Any, …)`), and on 1.13 those CodeInstances persist into the package image. After serialization the method and binding worlds collapse to the image-load world, so every other context (runtime, downstream precompilation) expands the generator fine — the poison is only ever *created* during our own precompile, and only *revived* for users whose device matches the workload's baseline target (`gfx1030` + wavefront 32, i.e. real RDNA2 users; CI's other targets never see it).

**Fix.** Move `struct KernelState` above the device include. Verified on Julia 1.13.0: the previously deterministic gfx1030 failure compiles cleanly and the emitted runtime library contains a plain `julia.gpu.state_getter` call with no dynamic dispatch.

**Guard.** The precompile workload now asserts that `kernel_state()` infers to `KernelState`, so this failure mode is a loud precompile error instead of a silently poisoned image (verified: reverting the reorder fails precompilation with `kernel_state() inferred as Any during precompilation`).

Complementary to JuliaGPU/GPUCompiler.jl#933, which partitions runtime-function caching away from the precompilation boundary so no backend can ship this class of poison again. Follow-ups worth considering: upstream, `get_staged` discarding real generator errors deserves at least a debug-visible path; and with the caching made safe, the workload could detect the host GPU to warm the right target.

https://claude.ai/code/session_01FJL9JQhcpXUVWkhuuiA7ad


simeonschaub (approved):

src/precompile.jl
```diff
@@ -33,6 +33,17 @@ if :AMDGPU in LLVM.backends()
             config = GPUCompiler.CompilerConfig(target, params;
                 kernel=true, name=nothing, always_inline=true)
 
+            # Generators run in their defining world; if `kernel_state`'s expansion fails
+            # here, `get_staged` swallows the error and inference caches `Any`-typed
+            # runtime code into our package image, miscompiling every bounds-checked
+            # kernel for users whose GPU matches this baseline target.
+            let mi = GPUCompiler.methodinstance(typeof(Device.kernel_state), Tuple{}),
```
vchuravy, line 40: I am not sure this is worth it, ideally we should just have a test

luraess: And possibly unfable (shorten/strip) the inline comment :-)

luraess: @gbaraldi would adding Valentin's suggestion and unfable comments be something you'd be up to so we can merge?

vchuravy (approved):

PR #1085, opened 2026-09-18
Fix Slurm account on CSCS

luraess:


luraess:
Fix done in the online config instead

PR #1084, opened 2026-09-18
Skip precompile workload on Windows to prevent hang

ehgus:
Fixes #1083

### Summary

In #1071, kernel linking was switched to in-process linking via `libamdgpu.AMDGPULink` (`AMDGPU_LLVM_Backend_jll` v23). 

Following this change, package precompilation hangs indefinitely on Windows during the warmup kernel compilation step. 

This PR skips the `@compile_workload` block on Windows (`if :AMDGPU in LLVM.backends() && !Sys.iswindows()`) as an immediate workaround to unblock Windows users.

PR #1087, opened 2026-09-18
Workaround precompile hang on Windows

### LINKED ISSUE #1083: BUG: `using AMDGPU` hangs indefinitely during precompilation on Windows on main branch
## Questionnaire

1. Does ROCm works for you outside of Julia, e.g. C/C++/Python?: Yes

2. Post output of `rocminfo`.
```
N/A (AMD ROCm on Windows does not provide the `rocminfo` utility).
```

3. Post output of `AMDGPU.versioninfo()` if possible.
```
# This is printed in v2.8.0
┌───────────┬──────────────────┬───────────┬───────────────────────────────────────────
│ Available │ Name             │ Version   │ Path                                     ⋯
├───────────┼──────────────────┼───────────┼───────────────────────────────────────────
│     +     │ LLD              │ -         │ C:\\Program Files\\AMD\\ROCm\\7.2\\bin\\ ⋯
│     +     │ Device Libraries │ -         │ C:\\Users\\MASKED\\.julia\\artif ⋯
│     +     │ HIP              │ 7.2.60201 │ C:\\Program Files\\AMD\\ROCm\\7.2\\bin\\ ⋯
│     +     │ rocBLAS          │ 5.2.0     │ C:\\Program Files\\AMD\\ROCm\\7.2\\bin\\ ⋯
│     +     │ rocSOLVER        │ 3.32.0    │ C:\\Program Files\\AMD\\ROCm\\7.2\\bin\\ ⋯
│     +     │ rocSPARSE        │ 4.2.0     │ C:\\Program Files\\AMD\\ROCm\\7.2\\bin\\ ⋯
│     +     │ rocRAND          │ 4.2.0     │ C:\\Program Files\\AMD\\ROCm\\7.2\\bin\\ ⋯
│     +     │ rocFFT           │ 1.0.36    │ C:\\Program Files\\AMD\\ROCm\\7.2\\bin\\ ⋯
│     -     │ hipTENSOR        │ -         │ -                                        ⋯
│     -     │ MIOpen           │ -         │ -                                        ⋯
└───────────┴──────────────────┴───────────┴───────────────────────────────────────────
                                                                       1 column omitted

AMDGPU devices
┌────┬────────────────────────┬──────────┬───────────┬────────────┬───────────────┐
│ Id │                   Name │ GCN arch │ Wavefront │     Memory │ Shared Memory │
├────┼────────────────────────┼──────────┼───────────┼────────────┼───────────────┤
│  1 │ AMD Radeon RX 7900 XTX │  gfx1100 │        32 │ 23.984 GiB │    64.000 KiB │
└────┴────────────────────────┴──────────┴───────────┴────────────┴───────────────┘```
```

## Reproducing the bug

1. Describe what's not working.

On Windows (Windows 11, Julia 1.13.0, ROCm 7.2), using AMDGPU (or Pkg.precompile()) on the main branch hangs indefinitely during precompilation and never finishes.

- On release v2.8.0 and on main branch just before the problematic commit (commit b3e831ef): Precompilation successfully completes.
- On main branch (commit 63c6c404): Precompilation hangs indefinitely with a frozen background Julia worker process:
```Julia REPL
(@v1.13) pkg> up
    Updating registry at `C:\Users\MASKED\.julia\registries\General.toml`
 Installed StaticArrays ─ v1.9.22
   Project No packages added to or removed from `C:\Users\MASKED\.julia\environments\v1.13\Project.toml`
  Updating `C:\Users\MASKED\.julia\environments\v1.13\Manifest.toml`
⌃ [61eb1bfa] ↑ GPUCompiler v2.6.0 ⇒ v2.7.0
  [90137ffa] ↑ StaticArrays v1.9.21 ⇒ v1.9.22
  [cc5c0156] ↑ AMDGPU_LLVM_Backend_jll v22.1.8+2 ⇒ v23.1.1+1
⌃ [f52de702] + LLVMDowngrader_jll v0.10.0+1
  [873c0968] - ROCmDeviceLibs_jll v7.0.2+0
        Info Packages marked with ⌃ have new versions available and may be upgradable.
Precompiling packages  ──────────────╸─────────  11/12
  ◐ AMDGPU
```

I could not find the root cause. 

2. Provide MWE to reproduce it (if possible).
```
]precompile AMDGPU
```

luraess:
Fixes #1083. Alternative to #1084.

#1071 moved kernel linking in-process via `libamdgpu.AMDGPULink`. Since then, precompilation hangs indefinitely on Windows.

@ehgus narrowed this down very precisely in #1083:

- `dlopen` of `libamdgpu` and `dlsym` of `AMDGPULink` are instant —> not a DLL-load problem.
- Forcing the in-process linker at **run time** (`lld_artifact = true`) links a kernel fine —> `AMDGPULink` is not broken on Windows.
- The stuck precompile worker sits at ~0% CPU, and merely inspecting its thread stack from Process Explorer releases the wait.

So the deadlock seems specific to running `AMDGPULink` inside a precompilation worker; it looks like a lost wakeup (`libamdgpu.dll` is a mingw-w64 build, so LLD's parallel task group runs on winpthreads condvars). That is an upstream problem, and `AMDGPULink` exposes no threading options to work around it.

### Fix

`create_executable` only reaches for the in-process linker during precompilation because ROCm discovery has not run yet, leaving `AMDGPU.lld_path` empty and not because the system lacks a linker. The reporter has `ld.lld.exe` in `C:\Program Files\AMD\ROCm\7.2\bin` the whole time.

The new `external_lld()` looks one up directly when generating package output, so precompilation links with the same `ld.lld` the runtime would use, and never enters `AMDGPULink`.

For the remaining case (Windows with no system `ld.lld`) the precompile workload skips just the warming compile, rather than the whole `@compile_workload` block as in #1084. That keeps the config/`methodinstance`/job construction warm, and mirrors the existing Julia < 1.12 carve-out directly above it.

Also switches the link call to `@gcsafe_ccall`, matching every other long-running foreign call in the package (not believed to be the root cause here).

### Testing

@ehgus, could you check this branch against #1083?


ehgus:
@luraess  This version is successfully precompiled!

luraess:
cscs-ci run

